PR #29, opened 2021-02-07
Feature/24 - Improved test case assertions

girardinsamuel:


tests/integrations/controllers/WelcomeController.py
```diff
@@ -11,3 +12,43 @@ def test(self):
 
     def view(self, view: View):
         return view.render("welcome")
+
+    def create(self, view: View):
```
girardinsamuel, line 16: @josephmancuso So here I am adding some controllers methods for the tests
and later in the test case I am adding those routes

tests/tests/test_testcase.py
```diff
@@ -1,20 +1,151 @@
...
+class User(Model, Authenticates):
+    pass
 
 
 class TestTesting(TestCase):
     def setUp(self):
         super().setUp()
-        self.addRoutes(Route.get("/test", "WelcomeController@show").name("test"))
+        self.addRoutes(
+            Route.get("/test", "WelcomeController@show").name("test"),
```
girardinsamuel, line 17: Here I am adding the test routes, but when I run tests the response for each route is always `b'welcome'`

src/masonite/tests/HttpTestResponse.py
```diff
@@ -37,8 +52,221 @@ def assertIsNotNamed(self, name=None):
...
+
+    def assertAuthenticatedAs(self, user):
+        user = self.application.make("auth").guard("web").user()
+        assert user == user
+        return self
+
+    def assertHasHttpMiddleware(self, middleware):
+        """Assert that the request/response cycle has the given middleware. The HTTP middleware
+        class should be given."""
+        assert middleware in self.application.make("middleware").http_middleware
```
girardinsamuel, line 197: @josephmancuso  not sure here if we should make one assertion method for all middlewares ?
Also middleware are registered with class in a list for HTTP and with a dict for routes one, so not sure what the user wants to check for now it would be
```python
assertHasHttpMiddleware(MyCustomMiddleware)
assertHasRouteMiddleware("web")
```
I could use * operator to be able to do assertHasRouteMiddleware("web", "auth") 🤔  ?

src/masonite/tests/HttpTestResponse.py
```diff
@@ -37,8 +52,221 @@ def assertIsNotNamed(self, name=None):
...
+            assert self.route.controller_class == controller
+        else:
+            raise ValueError("controller must be a string like YourController@index or a Controller class")
+        return self
+
+    def assertRouteHasParameter(self, key, value=None):
+        assert key in self.route.url_list, "Route does not contain parameter {key}."
+        if value is not None:
+            # TODO
+            # @josephmancuso not sure how to check if the route parameter has the given value
```
girardinsamuel, line 221: here maybe for now I can remove the value check as I don"t see a solution and maybe it's not worth it. We could use `assertLocation` here to check the full path

girardinsamuel:
PR at 80/90% I would say, I need to fix the tests and finish the JSON assertions.
I tried to add testing parity with masonite (but I renamed some of the assertions) and I added some other assertions from Laravel ones 😇

josephmancuso:
@girardinsamuel i fixed the code issues i think. Lot of tests are still failing but now it looks like normal issues like templates no found.

Also i wasn't sure what you were trying to do here. This is not valid Masonite code:

```python
    def create(self, view: View):
        return view("welcome", status=201)
```

Valid code would be:

```python
    def create(self, view: View):
        return view.render("welcome"), 201
```

josephmancuso:
Got you down to 9 failing tests that just need to be worked through ..

girardinsamuel:
@josephmancuso 
thanks for your help, I will fix tests and finish the PR soon. I already fixed some tests, down to 3 failing tests. I have an issue with `WebGuard` in assertAuthenticated, request is not bind to the application

tests/integrations/controllers/WelcomeController.py
```diff
@@ -14,27 +14,24 @@ def test(self):
...
         return "not found", 404
 
-    def unauthorized(self, view: View):
+    def unauthorized(self):
         return "unauthorized", 401
 
-    def forbidden(self, view: View):
-        return view.render("forbidden"), 403
+    def forbidden(self):
+        return "forbidden", 403
```
vaibhavmule, line 27: love this new response. @josephmancuso

josephmancuso: yes very clean :)

josephmancuso: This s actually possible in Masonite 3 too lol

girardinsamuel:
@josephmancuso  PR ready for review 👍 !

- I fixed some small bugs in the WebGuard/Authenticates classes 😉
- I fixed tests

there is this assertion `assertRouteHasParameter()` which right now can check if a route has given parameter name,
but I don't know how to check the related value (cf comments in the code) could you have a look at it ?
If it's too complex for now we could let this as comment and implement that later when improving some of the assertions.

src/masonite/tests/HttpTestResponse.py
josephmancuso: Yeah thats a good idea. Can always convert both to tuples and to issubset to check for the intersections between the 2 tuples. List might also have this support 

https://www.programiz.com/python-programming/methods/set/issubset

josephmancuso:
```python
    def assertHasHttpMiddleware(self, *middlewares):
        """Assert that the request/response cycle has the given middleware. The HTTP middleware
        class should be given."""
        assert set(middlewares).issubset(set(self.application.make("middleware").http_middleware)) 
```

something like this. Idk if it works

src/masonite/tests/HttpTestResponse.py
josephmancuso: Hmm yeah not really possible right now I don't think. Maybe we can set it on the route class when we are compiling the route ..?

src/masonite/response/response.py
```diff
@@ -167,38 +167,38 @@ def view(self, view, status=200):
...
         Keyword Arguments:
             location {string} -- The URL to redirect to (default: {None})
             status {int} -- The Response status code. (default: {302})
+            params {dict} -- The route params (default: {})
 
         Returns:
             string -- Returns the data to be returned.
         """
         self.status(status)
-        if params is None:
```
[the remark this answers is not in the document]
josephmancuso: So i think we should be good but the params = {} in the parameter list apparently is terrible Python code. The reason being is that the params are mutable. Meaning if you call this method multable times.

As I am typing this out tho i realized that dictionary are not mutable so maybe it doesn't apply to dictionaries .. lol 

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/masonite/routes/HTTPRoute.py
```diff
@@ -141,9 +141,6 @@ def get_response(self, app=None):
                 # *self.request.url_params.values() TODO
             )
 
-            if hasattr(response, "rendered_template"):
-                response = response.rendered_template
-
```
josephmancuso, line 146: Yeah we don't need this anymore cuss we added get_response() in the view class which gets called in the response class. This is an example of some old legacy code still in the codebase

josephmancuso (changes requested):

tests/integrations/controllers/WelcomeController.py
```diff
@@ -11,3 +13,85 @@ def test(self):
...
+        request.app.make("session").driver("cookie").flash("errors", {
+            "email": "Email required",
+            "password": "Password too short",
+            "name": ""
+        })
+        return "session"
+
+    def session2(self, request: Request):
+        # @josephmancuso: this should work, but in tests session is not bind to request
+        # request.session.flash("key", "value")
```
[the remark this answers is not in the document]
girardinsamuel: Looks like it's already done but inside tests I guess it's not called ?

josephmancuso (approved):

PR #39, opened 2021-02-11
Add way to mock services

girardinsamuel:
I tried an other approach here but still not sold. I am playing with this for now.

## Requirements
Let's clarify requirements (that I have in my mind)
- service and its mockable version should be configured somewhere so: 
  1. service could be mocked with an other mock implementation (e.g. from a package)
  2. we can retrieve the mock class to use to mock the given service OR we can retrieve the original service to bind back (to implement restore feature)
- M4 should provide a default implement of mocks for different services (mails, queues...) => easy
- a package should be able to update the mock implementation of a service.
- in the testscase we should be able to tell mock the given service : in a test or at global test class level.
- we should be able to start mocking / stop mocking: because maybe we want to write some integrations tests where we really want to send an email or test a service in real.

## Try 1
I also tried to add register_mocks() in Kernel to register all mock class of all services but then I don't know if an external package can update this and I don't know how to fetch the mock class from the test case.

## Try 2
I also tried to add a register_mock() in the service provider: so each service providers "knows how to mock himself" and mock implementation could be implemented easily by overriding the service provider. But I did not really manage to actually implement this.

## Try 3
this PR


girardinsamuel:
@josephmancuso What do you think of this ?
I would like to find a modular and powerful way to do this but without making services inherit from a base class (like `Mockable`) it's a bit harder.

josephmancuso:
Yeah i think just overriding the class with a mock class is the way to go. This would also allow us to make a lot of custom test assertions

girardinsamuel:
> Yeah i think just overriding the class with a mock class is the way to go. This would also allow us to make a lot of custom test assertions

So that's what I have done in this PR.
- I cleaned it a bit to make this "restorable" to be able to cancel the mock by reset class to the original.
- This will by default mock with the configured mock that we have to implement for the core services which needs it
- We can provide a custom mock class to the fake method to override mocking locally

tests/TestCase.py
```diff
@@ -106,6 +108,30 @@ def fetch(self, route, data=None, method=None):
...
         pass
 
+    def register_mocks(self):
+        """Configure the default mock classes for all services which needs to be mocked.
+        The mocks can now be overriden.
+        A package could in its service provider call a method which could update one of the mock
+        class with the one installed.
+        """
+        # here we will configure all default available mocks
+        self.application.bind("mock.mail", "src.masonite.tests.mocks.MockMail")
```
josephmancuso: Why do this rather than just bind the mock class directly?

josephmancuso: I changed this to bind to the container directly in the service provider. This will allow us to have packages that easily binds their mocks to the container

tests/features/mail/test_mock_mail.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestSMTPDriver(TestCase):
+    def test_mock_mail(self):
+        self.fake("mail")
+        welcome_email = self.application.make("mail").mailable(Welcome()).send()
+        (
+            welcome_email.seeEmailCc(None)
+            .seeEmailBcc(None)
+            .seeEmailContains("Hello from Masonite!")
+            .seeEmailContains("text from Masonite!")
```
josephmancuso, line 24: @girardinsamuel See the new email assertions that can now be done on the MockMail class

josephmancuso (approved):

josephmancuso (approved):

josephmancuso:
Made a few changes:

* Changed a bit how the mocks are registered (moved them from the testcase to a service provider and prefixed mock classes with `mock.`) This way any package can register mocks via their own service providers
* Mock classes now need to be registered via their class instance. This is so we can create a new instance everytime we call `self.fake('mail')`.
* Added tests for sending email to really test out this structure. Worked out great.

PR #41, opened 2021-02-13
feature/21 - Masonite Scheduling

josephmancuso:
Merged the Masonite-scheduler package into this codebase. Actually only took literally 5 minutes.

We will abandon that masonite-scheduler repo https://github.com/MasoniteFramework/scheduler when Masonite 4 comes out.

This PR still needs:

- [x] Clean codebase and improve upon from original task scheduling codebase
- [ ] Look at any additional features this could support

It would be cool if we could run a lambda function. Something like this:

```python
self.schedule(lambda: (
    self.application.make('mail').mailable(CronEmail(['joe@masoniteproject.com'])).send()
)).every('3 days')
```

## New Documentation

This PR adds the task scheduling from the masonite-scheduler repo to the core codebase.

This also adds a new Task Handler class (in keeping with the white paper we set out at the beginning of the project)

This handler class is really just an internal class but we can add tasks to it like we do commands and other things:

```python
self.appllication.make('scheduler').add(Task1(), Task2())
```

There is a wrapper around this if you make a service provider:

```python
from masonite.scheduling import CanSchedule

class ConsoleProvider(Provider, CanSchedule):

    def register(self):
         self.schedule(Task1()).daily_at(13,17) # runs daily at 1pm and 5pm
```

Theres a bunch of diff timing methods in the docs and thats really all there is to this PR ...

### DIFF
--- a/src/masonite/scheduling/CanSchedule.py
+++ b/src/masonite/scheduling/CanSchedule.py
@@ -1,0 +1,10 @@
+class CanSchedule:
+    def call(self, command):
+        command_class = CommandTask(command)
+        self.app.simple(command_class)
+        return command_class
+
+    def schedule(self, task):
+        task_class = task
+        self.app.simple(task_class)
+        return task_class
--- a/src/masonite/scheduling/CanSchedule.py
+++ b/src/masonite/scheduling/CanSchedule.py
@@ -1,10 +1,10 @@
 class CanSchedule:
     def call(self, command):
         command_class = CommandTask(command)
-        self.app.simple(command_class)
+        self.app.make("scheduler").add(command_class)
         return command_class
 
     def schedule(self, task):
         task_class = task
-        self.app.simple(task_class)
+        self.app.make("scheduler").add(task_class)
         return task_class
--- a/src/masonite/scheduling/Task.py
+++ b/src/masonite/scheduling/Task.py
@@ -125,8 +125,7 @@ class Task:
             return self._verify_run_at()
 
         if self.run_every_minute:
-            if self._date.minute / 1 == 1:
-                return True
+            return True
         elif self.run_every_hour:
             if self._date.hour / 1 == 1:
                 return True
--- a/src/masonite/scheduling/Task.py
+++ b/src/masonite/scheduling/Task.py
@@ -1,0 +1,161 @@
+import pendulum
+
+
+class Task:
+
+    run_every = False
+    run_at = False
+    run_every_hour = False
+    run_every_minute = False
+    twice_daily = False
+    run_weekly = False
+
+    _date = None
+
+    name = ""
+
+    def __init__(self):
+        """
+        Should only be on the child class. Also needs to be resolved by the container.
+        """
+
+        pass
+
+    def every(self, time):
+        self.run_every = time
+        return self
+
+    def every_minute(self):
+        self.run_every = "1 minute"
+        return self
+
+    def every_15_minutes(self):
+        self.run_every = "15 minutes"
+        return self
+
+    def every_30_minutes(self):
+        self.run_every = "15 minutes"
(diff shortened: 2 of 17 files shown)


src/masonite/scheduling/CanSchedule.py
```diff
@@ -0,0 +1,10 @@
+class CanSchedule:
+    def call(self, command):
+        command_class = CommandTask(command)
```
girardinsamuel, line 3: Should not this class be imported ?

src/masonite/scheduling/Task.py
```diff
@@ -0,0 +1,147 @@
...
+    def every_minute(self):
+        self.run_every = "1 minute"
+        return self
+
+    def every_15_minutes(self):
+        self.run_every = "15 minutes"
+        return self
+
+    def every_30_minutes(self):
+        self.run_every = "15 minutes"
```
girardinsamuel, line 37:
```suggestion
        self.run_every = "30 minutes"
```

src/masonite/scheduling/Task.py
```diff
@@ -0,0 +1,147 @@
...
+        return self
+
+    def every_45_minutes(self):
+        self.run_every = "15 minutes"
```
girardinsamuel, line 41:
```suggestion
        self.run_every = "45 minutes"
```

tests/features/scheduling/test_scheduling.py
```diff
@@ -0,0 +1,160 @@
...
+        time = pendulum.now().on(2018, 1, 1).at(12, 20, 5)
+        self.task.run_every = ""
+        self.task.twice_daily = (1, 13)
+        self.task._date = time
+
+        assert self.task.should_run() is False
+
+    def test_run_at(self):
+        self.task.run_every = ""
+        self.task.run_at = None
```
girardinsamuel, line 130: Why those two lines to set same attribute ?

tests/features/scheduling/test_scheduling.py
```diff
@@ -0,0 +1,160 @@
...
+        self.task.run_at = "13:00"
+
+        time = pendulum.now().on(2018, 1, 1).at(13, 0, 5)
```
girardinsamuel, line 133: I am not sure of what you're trying to do here. Are some asserts missing ?

tests/features/scheduling/test_scheduling.py
```diff
@@ -0,0 +1,160 @@
...
+        time = pendulum.now().on(2018, 1, 1).at(13, 6, 5)
+        self.task._date = time
+
+        assert self.task.should_run() is False
+
+    def test_method_calls(self):
+        task = MockTask()
+        task.at("13:00")
+
+        time = pendulum.now().on(2018, 1, 1).at(13, 0, 5)
```
girardinsamuel, line 152: What's the goal of two lines here ? There is no call

src/masonite/scheduling/Task.py
```diff
@@ -0,0 +1,147 @@
...
+
+        pass
+
+    def should_run(self, date=None):
+        """If the task should run"""
+
+        # set the date
+        self._set_date()
+
+        if self._verify_run():
```
girardinsamuel, line 78: Maybe here you could return `self._verify_run()` directly ?

girardinsamuel (changes requested):
The API looks nice 👍🏼

src/masonite/scheduling/Task.py
josephmancuso: I like it

tests/features/scheduling/test_scheduling.py
josephmancuso: just setting the time to 1pm

PR #42, opened 2021-02-14
Add location helpers

girardinsamuel:
- [x] views
- [x] controllers
- [x] seed files
- [x] migrations files
- [x] config

We could also add but those could be really more project specific and I am not sure we could enforce good conventions on those so if we choose to finally add those helpers, guess we can start with the ones above.
- models 
- resources folder

### DIFF
--- a/src/masonite/foundation/Kernel.py
+++ b/src/masonite/foundation/Kernel.py
@@ -39,6 +39,7 @@ class Kernel:
         self.application.bind("config.session", "tests.integrations.config.session")
         self.application.bind("config.queue", "tests.integrations.config.queue")
         self.application.bind("config.database", "tests.integrations.config.database")
+        self.application.bind("config.location", "tests/integrations/config")
 
     def register_controllers(self):
         self.application.bind("controller.location", "tests.integrations.controllers")
@@ -57,6 +58,8 @@ class Kernel:
                 ).DATABASES
             ),
         )
+        self.application.bind("migrations.location", "tests/integrations/databases/migrations")
+        self.application.bind("seeds.location", "tests/integrations/databases/seeds")
 
     def register_storage(self):
         storage = StorageCapsule(self.application.base_path)
--- a/src/masonite/utils/location.py
+++ b/src/masonite/utils/location.py
@@ -3,17 +3,39 @@ location."""
 from os.path import join, abspath
 
 
+def _normalize_path(path):
+    """Transform dotted path to normal file path."""
+    return path.replace(".", "/")
+
+def _build_path(location_key, relative_path, absolute):
+    from wsgi import application
+    relative_dir = join(_normalize_path(application.make(location_key)), relative_path)
+    return abspath(relative_dir) if absolute else relative_dir
+
 def view_path(relative_path, absolute=True):
     """Build the absolute path to the given relative_path assuming it exists in the configured
     views location. The relative path can be returned instead by setting absolute=False."""
-    from wsgi import application
-    relative_dir = join(application.make("views.location"), relative_path)
-    return abspath(relative_dir) if absolute else relative_dir
-
+    return _build_path("views.location", relative_path, absolute)
 
 def controller_path(relative_path, absolute=True):
     """Build the absolute path to the given relative_path assuming it exists in the configured
     controllers location. The relative path can be returned instead by setting absolute=False."""
-    from wsgi import application
-    relative_dir = join(application.make("controller.location"), relative_path)
-    return abspath(relative_dir) if absolute else relative_dir
+    return _build_path("controller.location", relative_path, absolute)
+
+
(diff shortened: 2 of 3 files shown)


src/masonite/foundation/Kernel.py
```diff
@@ -39,6 +39,7 @@ def set_framework_options(self):
         self.application.bind("config.session", "tests.integrations.config.session")
         self.application.bind("config.queue", "tests.integrations.config.queue")
         self.application.bind("config.database", "tests.integrations.config.database")
+        self.application.bind("config.location", "tests/integrations/config")
```
josephmancuso, line 42: How will this be used?

src/masonite/utils/location.py
```diff
@@ -0,0 +1,44 @@
...
+from os.path import join, abspath
+
+
+def _normalize_path(path):
+    """Transform dotted path to normal file path."""
+    return path.replace(".", "/")
+
+
+def _build_path(location_key, relative_path, absolute):
+    from wsgi import application
```
josephmancuso, line 12: should we hard code this here? I'm hesitant on hard coding location paths on something that is used by the developer

girardinsamuel: I am not sure of you mean by this. the idea behind this PR is to be avoid hard-coding paths, ensure consistency throughout the codebase and avoid hard-coding path building
I will clarify the idea and how/when to use this in next comment 😉

girardinsamuel:
Use cases are when we want to:
- to publish migrations from a package to the project migrations directory or publish any other resources from a package
- to scaffold a new controller in the project directory (craft new controller)
- to scaffold any other resources ..

With M4 you wanted to avoid hard-coding location of controllers, views and others ... in the project

So the idea is to let user configure/update the location of the different resources through the Kernel which is configurable per project (if needed) as I understood.

then in the scaffolding commands or in the external packages scaffolding we will always need to build the output paths for the stub resource we want to copy over:

**Example 1*
in the ControllerCommand used to scaffold a new controller we would now want the controller to be scaffolded in the controllers directory configured from the kernel and then build the path to it so we can copy over the stub controller.
```python
# M3 code
    def handle(self):
        controller = self.argument("name")
        view = View(App())

        if not self.option("exact"):
            controller = controller + "Controller"

        if not make_directory("app/http/controllers/{0}.py".format(controller)):  # <-- here
            return self.line_error("{0} Controller Exists!".format(controller))

        with open("app/http/controllers/{0}.py".format(controller), "w+") as f: # <-- here
            # (.....)
        self.info("Controller Created Successfully!")
```

`app/http/controllers/{0}.py"` is hard-coded in several place and also the building of the path to resource we want to copy, here I propose to use
```python
output_controller = controller_path("{controller}.py")
```
and just like this we have the output path built accordingly to the configured directory.

**Example 2**
In a package service provider we will need to copy migrations, views, controllers... over to the project where the package is used. And we could easily use those helpers to get the output path of those resources (without rebuilding the path manually and ensuring that output path is correct regarding the project config).
This way future masonite packages which will use those will ensure consistency with the paths they use and they won't bother creating the paths manually for all the resources they have to copy over.

And if we want to build the path ourself to have total control, well we are not obliged to use those helpers.

This would assume (as I did in this PR) that we would configure all resources default path inside Kernel as you started doing e.g. for views and that those helpers will use those values (which can be overriden by the users).

src/masonite/foundation/Kernel.py
girardinsamuel: For the config the idea is not hard-code the "config" directory.
For example with `masonite-inertia` when I want to publish the config file, I can use `config_path("inertia.py")` to compute the  inertia config output path in the project and it's easier to copy it over. (the user in his project might define an other config directory).

Not the best example as that might be the directory which is less likely to be changed but it's an example

girardinsamuel:
It's like the Laravel "paths" helpers
https://laravel.com/docs/8.x/helpers#paths

src/masonite/foundation/Kernel.py
josephmancuso: gotcha ok tha works

src/masonite/utils/location.py
josephmancuso: Right but what if the application is not in the wsgi.py file? should we require at least this file?

girardinsamuel: ah..was thinking that this was always there

If we look at the dynamic configuration of views in here
```python
class Kernel:
    # ....
    def register_templates(self):
        self.application.bind("views.location", "tests/integrations/templates")
```
This makes sense and can be used only if there is an application with this Kernel registered. As is I feel that the wsgi.py file defined as it is now is already a requirement for the project, but maybe that's the only one as almost all other parts are built dynamically and can be swapped out or removed.

josephmancuso (approved):

PR #43, opened 2021-02-15
Added Caching Feature

### LINKED ISSUE #23: Add Caching

josephmancuso:
Closes #23 

This adds a caching feature. This has multiple "Stores" which are like database connections. To add something to the default store you would do something like this:

```python
def show(self, cache: Cache):
    cache.store().put(key, value, seconds)
```

^^ Not 100% sold on this as it seems like a needless call but one of the few downsides to this structure is that managers are singletons now which means if you switch a driver in 1 part of the request like in a middleware, it will remain switched forever until you switch it back .. With the above approach at least we create a new driver class when this is called so we get around it and its not THAT big of a deal ..

To use a specific store:

```python
def show(self, cache: Cache):
    cache.store('sessions').put(key, value, seconds)
```

### Todo:

- [x] `redis` driver. (honestly did not know we had this 😓 )
- [x] `memcache` driver
- [x] `add` method (get or create method)

### DIFF
--- a/src/masonite/cache/drivers/FileDriver.py
+++ b/src/masonite/cache/drivers/FileDriver.py
@@ -21,7 +21,7 @@ class FileDriver:
             return self.get(key)
 
         self.put(key, value)
-        return self.get(key)
+        return value
 
     def get(self, key, default=None, **options):
         if not self.has(key):
@@ -87,7 +87,7 @@ class FileDriver:
 
     def get_expiration_time(self, seconds):
         if seconds is None:
-            seconds = 9999999
+            seconds = 31557600 * 10
 
         return seconds
 
--- a/src/masonite/cache/drivers/FileDriver.py
+++ b/src/masonite/cache/drivers/FileDriver.py
@@ -34,7 +34,7 @@ class FileDriver:
 
             if modified_at.add(seconds=self.get_cache_expiration(value)).is_past():
                 self.forget(key)
-                return None
+                return default
 
             value = self.get_value(value)
 
--- a/src/masonite/cache/drivers/FileDriver.py
+++ b/src/masonite/cache/drivers/FileDriver.py
@@ -16,6 +16,13 @@ class FileDriver:
             make_full_directory(options.get("location"))
         return self
 
+    def add(self, key, value):
+        if self.has(key):
+            return self.get(key)
+
+        self.put(key, value)
+        return self.get(key)
+
     def get(self, key, default=None, **options):
         if not self.has(key):
             return None
--- a/src/masonite/cache/drivers/FileDriver.py
+++ b/src/masonite/cache/drivers/FileDriver.py
@@ -1,0 +1,97 @@
+import os
+from ...utils.filesystem import make_full_directory, modified_date
+from pathlib import Path
+import pendulum
+import json
+import glob
+
+
+class FileDriver:
+    def __init__(self, application):
+        self.application = application
+
+    def set_options(self, options):
+        self.options = options
+        if options.get("location"):
+            make_full_directory(options.get("location"))
+        return self
+
+    def get_modified_at(self, filename):
+        return pendulum.from_timestamp(modified_date(filename))
+
+    def get(self, key, default=None, **options):
+        if not self.has(key):
+            return None
+
+        modified_at = self.get_modified_at(os.path.join(self._get_directory(), key))
+
+        with open(os.path.join(self._get_directory(), key), "r") as f:
+            value = f.read()
+
+            if modified_at.add(seconds=self.get_cache_expiration(value)).is_past():
+                self.forget(key)
+                return None
+
+            value = self.get_value(value)
+
+        return value
+
+    def get_expiration_time(self, seconds):
+        if seconds is None:
+            seconds = 9999999
+
+        return seconds
+
+    def get_value(self, value):
+        value = str(value.split(":", 1)[1])
+        if value.isdigit():
+            return str(value)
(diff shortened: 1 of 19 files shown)


girardinsamuel:
Should we also implement `memcached` ? It's the "default go-to" key value store that is used with Django.
From their docs:

> The fastest, most efficient type of cache supported natively by Django, Memcached is an entirely memory-based cache server, originally developed to handle high loads at LiveJournal.com and subsequently open-sourced by Danga Interactive. It is used by sites such as Facebook and Wikipedia to reduce database access and dramatically increase site performance.
Memcached runs as a daemon and is allotted a specified amount of RAM. All it does is provide a fast interface for adding, retrieving and deleting data in the cache. All data is stored directly in memory, so there’s no overhead of database or filesystem usage.
After installing Memcached itself, you’ll need to install a Memcached binding. There are several Python Memcached bindings available; the two most common are python-memcached and pylibmc.

josephmancuso:
@girardinsamuel considering I just added redis support in 20 min i think we should deff explore memached.

Right now all the tests are failing because I don't have a redis server connected

girardinsamuel:
For CI (not talking about local setup), ideally we could have two tests suites (just categorized with pytest markers):
- a quick one which should include most of the tests
- a "integrations tests" one which will be slower and which could need some additional services. For this one we could use additional setup in github action workflow to define required services. For example add redis https://github.com/marketplace/actions/redis-server-in-github-actions.

for the test suites it's just a matter of doing
```python
import pytest

def quick_test():
    pass 

@pytest.mark.integrations (or whatever keyword: slow...)
def slow_test():
    pass
```
and then running:
```
python -m pytest tests  # run all tests
python -m pytest tests -m"not integrations" # run quick tests
python -m pytest tests -m"integrations" # run integrations tests only
```

josephmancuso:
I think this is ready for a review. Its a rather simple feature. Currently now supports `file`, `redis` and `file` drivers. I think in the future and if the. need is there we can pretty simply support the `database` driver. Can revisit this in the future. Maybe on the "road to beta" project

src/masonite/cache/drivers/FileDriver.py
```diff
@@ -0,0 +1,104 @@
...
+        if options.get("location"):
+            make_full_directory(options.get("location"))
+        return self
+
+    def add(self, key, value):
+        if self.has(key):
+            return self.get(key)
+
+        self.put(key, value)
+        return self.get(key)
```
girardinsamuel, line 24: here why not return value directly ?

src/masonite/cache/drivers/MemcacheDriver.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.get_value(
+            self.get_connection().get(f"{self.get_name()}_cache_{key}")
+        )
+
+    def put(self, key, value, seconds=0, **options):
+        if isinstance(value, (dict, list)):
+            value = json.dumps(value)
+
+        return self.get_connection().set(
+            f"{self.get_name()}_cache_{key}", value, expire=seconds
```
girardinsamuel, line 51:
```python
f"{self.get_name()}_cache_{key}"
```
It's a convention you choose for the key ? To provide "unicity" ? (I mean instead of just using `key`

src/masonite/cache/drivers/MemcacheDriver.py
```diff
@@ -0,0 +1,94 @@
...
+        if value:
+            return value
+
+        callable(self)
+
+    def forget(self, key):
+        try:
+            self.get_connection().delete(f"{self.get_name()}_cache_{key}")
+            return True
+        except FileNotFoundError:
```
girardinsamuel, line 75: For memcache, is it really a FileNotFoundError exception ? 
When is it raised ? if the delete operation fail ? or the key not found ?

I am not an expert 😅 , but just feels weird

src/masonite/cache/drivers/RedisDriver.py
```diff
@@ -0,0 +1,99 @@
...
+        if value:
+            return value
+
+        callable(self)
+
+    def forget(self, key):
+        try:
+            self.get_connection().delete(f"{self.get_name()}_cache_{key}")
+            return True
+        except FileNotFoundError:
```
girardinsamuel, line 77: same remark here, I guess it's okay for FileDriver but feels weird for RedisDriver ?

src/masonite/cache/drivers/FileDriver.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _get_directory(self):
+        return self.options.get("location")
+
+    def get_modified_at(self, filename):
+        return pendulum.from_timestamp(modified_date(filename))
+
+    def get_expiration_time(self, seconds):
+        if seconds is None:
+            seconds = 9999999
```
girardinsamuel, line 90: this is the max expiration time ? 
For HTTP cache there are no conventions but we often see something like 1 year so (31557600 s)
Here the value is a bit less than 4 months, isn't it a bit short ? 

Or maybe this value should go in the config with a default of one year ?

girardinsamuel (changes requested):
Look really nice and clear ! 
I made some small remarks essentially

src/masonite/cache/drivers/MemcacheDriver.py
josephmancuso: Took it from the redis integration someone did for Masonite .. I think they did it because if you have multiple apps using the same memcache /redis server

PR #50, opened 2021-02-20
feature/49 - Adds Storage & Uploading

### LINKED ISSUE #49: Add Uploading support back (File system feature)
In Masonite M3 we can upload files to various places. Supported are local storage and Amazon S3.. 

The feature is extremely simple though and doesn't really support anything else.

We also have a `Storage` class which is a half-baked feature that tried to bridge the gap between storage and uploading.

I think we can take this and run with it and make it more similar to this Laravel https://laravel.com/docs/8.x/filesystem

josephmancuso:
Closes #49 

This PR adds a new `Storage` feature which is also adds all the uploading support, filesystem management, deleting, moving, copying, etc.

For uploading this will look like:

```python
def upload(self, storage: Storage):
    storage.disk().store(request.input('profile'))
    storage.disk().store(request.input('profile'), 'file-name')
```

All inputs that are files (or bytes) it will now be an instance of `UploadedFile` rather than a string or whatever the value is. This allows us to to do some cool things


josephmancuso (commented):
@girardinsamuel if you can give this a quick review. This hits all the requirements we have in M3 at least

PR #51, opened 2021-02-21
Add notifications

### LINKED ISSUE #32: Add notifications
Closes #51 

As discussed I guess I will migrate this PR https://github.com/MasoniteFramework/notifications/pulls to a PR in this repo

girardinsamuel:
Fixes #32

## Specs
- `mail` notification
- `slack` notifications
- `database` driver to save notifications in database
- `broadcast` driver to broadcast notifications
- `vonage` driver (ex Nexmo) to send SMS notifications
- add a `Notifiable` mixin to set app models notifiable (such as User).
- add a way to send notifications to not Notifiable entity aka AnonymousUsers
- be able to make notification quickly as we can make mailable
- built so that adding a new notification driver internally (or through an external package) can be done easily

## Documentation
Doc has been written here https://github.com/MasoniteFramework/docs/pull/131, so most of it could be taken from this PR when writing M4 docs 🚀 

### TLDR
We can create a notification with `craft notification Welcome`.
We can send it via the different drivers either by resolving `notification`:
```python
# send to anonymous entities (not a model)
self.application.make("notification")
    .route("mail", "sam@masoniteproject.com")
    .route("slack", "#general")
    .send(WelcomeNotification())
```
or by sending to a `Notifiable` entity (such as User):
```python
class User(Model, Notifiable):

     def route_notification_for_mail(self):
          return self.email
          # or return [self.pro_email, self.personal_email]
   
     def route_notification_for_slack(self):
          return self.user_channel_id
# ..
user.notify(WelcomeNotification())
```

A notification is built this way:
```python
class WelcomeNotification(Notification):
    def to_mail(self, notifiable):
         return Mailable().from_(..).text("Welcome !)
    
    def to_slack(self, notifiable):
         return SlackMessage().from("Sam").text("Welcome !)

    def via(self, notifiable):
         return ["mail", "slack"]
```
meaning that for each driver we should define a `to_{driver}` explaining how to build the notification. We have access to the notifiable here (e.g. our user instance for non anonymous notifications).
The `via()` explains through which drivers it should be sent. You can run any logic you want here because you have access to notifiable ! Of course when sending notification, it can be overriden :
```python
user.notify(drivers=[“mail"], WelcomeNotification())
```

If you want to queue the notification sending operation you can subclass the notification with `ShouldQueue`, that's it. For each driver, the sending job will be pushed onto the queue.

If you want to store notifications in database also, it's done as a driver also. You should first create the notifs table `craft notification:table`. The default name is `notifications` but can be overriden in the config. Then update your notification:
```python
class WelcomeNotification(Notification):
    # ...
    def to_database(self, notifiable):
          # define here what data will be serialized into database
          return {"data": "Welcome {notifiable.name}"}
    
    def via(self, notifiable):
         return ["mail", "slack", "database"]
```
then 
```python
user.notify(WelcomeNotification())
# and you can acccess
user.notifications.count() # == 1
user.read_notifications()
user.unread_notifications()
notification.mark_as_read()
notification.mark_as_unread()
```

Adding new notification channels/drivers is really easy, all we need is to create a new driver and register it. It should implement send() and queue() methods and can use config/notifications.py to store its own config.

## TODO
- [x] globally defined dry mode
- [x] notification mocks for testing as done for others
- [x] finish adding drivers
- [x] add integrations tests with notification to different channels
- [x] add command to create notification
- [x] ~~finish adding queuing for all drivers~~ (postponed)
- [x] find a solution to be able to override mail driver used in mail notification


src/masonite/drivers/notification/DatabaseDriver.py
```diff
@@ -0,0 +1,41 @@
+"""Database driver Class."""
+import json
+from masonite.app import App
+from masonite import Queue
+
+from ..models import DatabaseNotification
```
josephmancuso, line 6: Not sure of the issues this will cause but in other database drivers I just get the application and call `self.application.make('builder')`. Maybe we should do that? This calls the query builder that is bound in the kernel class

src/masonite/drivers/notification/MailDriver.py
```diff
@@ -0,0 +1,66 @@
+"""Mail driver Class."""
+
+from ...exceptions.exceptions import NotificationException
+from .BaseDriver import BaseDriver
+
+
+class MailDriver(BaseDriver):
```
josephmancuso, line 7: Maybe we need to use the `Mailable` class here to build the email notification ..

src/masonite/drivers/notification/vonage/Sms.py
```diff
@@ -0,0 +1,46 @@
+"""Vonage Sms Component"""
```
josephmancuso, line 1: This would be huge. I love having the idea of sending a verification code to users.

src/masonite/notification/Notification.py
```diff
@@ -0,0 +1,51 @@
+"""Base Notification facade."""
+from abc import ABC, abstractmethod
```
josephmancuso, line 2: I think we can remove this. Although having contract classes are great I haven't actually seen them be very useful. Especially because Python is not typed and this only ensures the methods are on the class and not necessarily that the signature of those methods match. So its like a half-ass contract class .. I just haven't seen them be useful at all. I haven't added them on any other features and I haven't ported them over from M3 .. 

I'm open for conversation on this but yeah I don't see it being a useful thing to add

src/masonite/notification/NotificationManager.py
```diff
@@ -0,0 +1,146 @@
...
+import uuid
+from masoniteorm.models import Model
+
+from ..utils.collections import Collection
+from ..drivers.notification.BaseDriver import BaseDriver
+from ..exceptions.exceptions import NotificationException, DriverNotFound
+from ..queues import ShouldQueue
+
+
+class NotificationManager(object):
```
josephmancuso, line 11: We don't have to call it `NotificationManager`. If we base off the other features I just did a simple class like `Notification`.

src/masonite/notification/Notification.py
girardinsamuel: Okay I'll remove

src/masonite/notification/NotificationManager.py
girardinsamuel: Actually I called it this way because I have a Notification class which represent a notification that can be sent

src/masonite/drivers/notification/MailDriver.py
girardinsamuel: I have used the Mailable class but not in the way you meant maybe. 
When you build the notification, the mail_to method should return a Mailable

src/masonite/drivers/notification/DatabaseDriver.py
girardinsamuel: Finally I am not sure of the impact of what you're suggesting,
Here I am importing the model, do you want that we don't import the model and use only table name like this :
```
self.application.make('builder').table("notifications").create(model_data)
# instead of 
from ..models import DatabaseNotification
DatabaseNotification.create(model_data)
``` 

I want to be as flexible as possible. This feature is needing a table ("notifications") and the related model Notification provided in the codebase. Maybe we should give the user ability to define its own table name ?

girardinsamuel: oh no okay I have seen what was done in QueueDatabaseDriver, really nice. I have done the same now

src/masonite/notification/NotificationManager.py
girardinsamuel: I can rename everything to send `Message` instead of `Notification` and then I can call the feature class `Notification` but I don't know if it's better. Everyone know what a Notification, I am not sure if we should call it another way. 
🤔

src/masonite/stubs/notification/Notification.py
```diff
@@ -0,0 +1,16 @@
...
+from masonite.mail import Mailable
+
+
+class __class__(Notification):
+    def to_mail(self, notifiable):
+        return (
+            Mailable()
+            .to(notifiable.email)
+            .subject("Masonite 4")
+            .from_("sam@masoniteproject.com")
```
josephmancuso, line 11: Make this more generic. Like `hello@email.com`

src/masonite/stubs/notification/create_notifications_table.py
```diff
@@ -0,0 +1,17 @@
+from masoniteorm.migrations import Migration
+
+
+class CreateNotificationsTable(Migration):
+    def up(self):
+        """Run the migrations."""
+        with self.schema.create("notifications") as table:
+            table.string("id", 36).primary()
```
josephmancuso, line 8: Usually there are a lot of notifications right? Should we default this to `big_increments`?

josephmancuso (changes requested):
1. One thing I noticed, I kinda of changed my idea on this after implementing the second or third feature is to make everything its own module.

So in short we should put something like the `VonageDriver` from

```diff
- src/masonite/drivers/notification/vonage/VonageDriver.py
+ src/masonite/notification/drivers/VonageDriver.py
```

This will put all things related to notifications inside its own module

2. We should add back the database.sqlite3 file. I think removing it will break a lot of tests. This is the same as the ORM repo where we have that one sqlite file we run integration tests on

src/masonite/stubs/notification/create_notifications_table.py
girardinsamuel: Yeah maybe, here I was using id as `uuid` but UUID column in ORM was not released at the time I wrote this line.
I propose to keep uuid but to use the real uuid column from ORM. Are you ok ?

girardinsamuel:
>     1. One thing I noticed, I kinda of changed my idea on this after implementing the second or third feature is to make everything its own module.

I made changes 

>  We should add back the database.sqlite3 file.
I did not removed it, I just edited it to add a phone field to user to demonstrate different use cases during tests...

src/masonite/notification/DatabaseNotification.py
```diff
@@ -0,0 +1,38 @@
+"""DatabaseNotification Model."""
+import pendulum
+from masoniteorm.relationships import morph_to
+from masoniteorm.models import Model
+
+
+class DatabaseNotification(Model):
```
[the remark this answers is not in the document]
girardinsamuel: Maybe we could but then all features should do this as right now there are other places where we are using ORM and it's tied with the feature...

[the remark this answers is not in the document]
girardinsamuel: masoniteorm is "sprinkled" in a lot of place in masonite4. Its query builder is abstracted with builder key in the container. But then some migrations and models are there. Developers are not obliged to use it but I guess we should at least provide e.g. a DatabaseNotification model ready to be used with default orm which is Masonite ORM.

Maybe the database notification driver is a ORMDatabaseDriver. If a developer want to use an other ORM, he will need to make a provider to swap the database driver with his custom one where he specifies how to save a notifcation in database and which model to use ... but Notifiable class should also be changed ...
it looks difficult !

josephmancuso:
@girardinsamuel let's ignore the queuing of notifications for now that can be a phase 2 thing. I wanna get this

josephmancuso:
@girardinsamuel I modified this a bit.

Now a notification class takes a basic `Notification` class it did before but that class now does really nothing but contains some of those utility methods (like should_send and ignore_errors).

Now a notification class must inherit from different utility classes to do things. To send an email it will inherit from the normal Mailable class to build a mailable.

It can also inherit from a `Textable` class to send those Vonage text messages.

This is an example of a notification class:

```python
class OneTimePassword(Notification, Mailable, Textable):

    def to_mail(self, notifiable):
        return (
            self
            .to(notifiable.email)
            .subject("Masonite 4")
            .from_("hello@email.com")
            .text(f"Hello {notifiable.name}")
        )

    def to_vonage(self, notifiable):
        return self.text_message("Welcome !").to('6314870798').from_("33123456789")

    def via(self, notifiable):
        return ["vonage"]
```

Notice it inherits from a view different classes

josephmancuso (changes requested):

src/masonite/notification/NotificationManager.py
```diff
@@ -0,0 +1,93 @@
...
+from ..exceptions.exceptions import NotificationException
+from ..queues import ShouldQueue
+from .AnonymousNotifiable import AnonymousNotifiable
+
+
+class NotificationManager:
+    """Notification handler which handle sending/queuing notifications anonymously
+    or to notifiables through different channels."""
+
+    # those classes are use for mock => should we put them only in NotificationMock ?
```
josephmancuso, line 13: remove

src/masonite/notification/NotificationManager.py
```diff
@@ -0,0 +1,93 @@
...
+                        results.append(driver_instance.queue(notifiable, notification))
+                    else:
+                        results.append(driver_instance.send(notifiable, notification))
+                except Exception as e:
+                    if not notification.ignore_errors() and not fail_silently:
+                        raise e
+
+        return results[0] if len(results) == 1 else results
+
+    # def is_custom_channel(self, channel):
```
josephmancuso, line 76: remove

src/masonite/notification/NotificationManager.py
```diff
@@ -0,0 +1,93 @@
...
+            return notifiables
+        else:
+            return [notifiables]
+
+    def route(self, driver, route):
+        """Specify how to send a notification to an anonymous notifiable."""
+        return AnonymousNotifiable(self.application).route(driver, route)
+
+    # TESTING
+    def assertNotificationDried(self, notification_class):
```
josephmancuso, line 92: remove?

src/masonite/notification/drivers/BaseDriver.py
```diff
@@ -0,0 +1,26 @@
+class BaseDriver:
+    def send(self, notifiable, notification):
+        """Implements sending the notification to notifiables through
+        this driver."""
+        raise NotImplementedError(
+            "send() method must be implemented for a notification driver."
+        )
+
+    def queue(self, notifiable, notification):
```
josephmancuso, line 9: remove for now

src/masonite/notification/drivers/BroadcastDriver.py
```diff
@@ -0,0 +1,31 @@
...
+    def send(self, notifiable, notification):
+        """Used to broadcast a notification."""
+        data = self.get_data("broadcast", notifiable, notification)
+        channels = notification.broadcast_on() or notifiable.route_notification_for(
+            "broadcast"
+        )
+        event = notification.type()
+        self.application.make("broadcast").channel(channels, event, data)
+
+    def queue(self, notifiable, notification):
```
josephmancuso, line 24: remove for now

src/masonite/notification/drivers/DatabaseDriver.py
```diff
@@ -0,0 +1,44 @@
...
+            .on(self.options.get("connection"))
+            .table(self.options.get("table"))
+        )
+
+    def send(self, notifiable, notification):
+        """Used to send the email and run the logic for sending emails."""
+        data = self.build(notifiable, notification)
+        return self.get_builder().new().create(data)
+
+    def queue(self, notifiable, notification):
```
josephmancuso, line 28: remove for now

src/masonite/notification/drivers/MailDriver.py
```diff
@@ -0,0 +1,33 @@
...
+    def send(self, notifiable, notification):
+        """Used to send the email."""
+        mailable = self.get_data("mail", notifiable, notification)
+        if not mailable._to:
+            recipients = notifiable.route_notification_for("mail")
+            mailable = mailable.to(recipients)
+        # TODO: allow changing driver how ?????
+        return self.application.make("mail").mailable(mailable).send(driver="terminal")
+
+    def queue(self, notifiable, notification):
```
josephmancuso, line 24: remove for now

src/masonite/notification/drivers/MailDriver.py
```diff
@@ -0,0 +1,33 @@
...
+        self.options = options
+        return self
+
+    def send(self, notifiable, notification):
+        """Used to send the email."""
+        mailable = self.get_data("mail", notifiable, notification)
+        if not mailable._to:
+            recipients = notifiable.route_notification_for("mail")
+            mailable = mailable.to(recipients)
+        # TODO: allow changing driver how ?????
```
josephmancuso, line 21: I think we can set this on the mailable right? Not sure if we have this now but we should be able to

src/masonite/notification/drivers/SlackDriver.py
```diff
@@ -0,0 +1,117 @@
...
+    def send(self, notifiable, notification):
+        """Used to send the notification to slack."""
+        slack_message = self.build(notifiable, notification)
+        if slack_message._mode == self.WEBHOOK_MODE:
+            self.send_via_webhook(slack_message)
+        else:
+            self.send_via_api(slack_message)
+
+    # def queue(self, notifiable, notification):
+    # TODO
```
josephmancuso, line 33: remove

src/masonite/notification/drivers/SlackDriver.py
```diff
@@ -0,0 +1,117 @@
...
+        """Build Slack message payload sent to Slack API or through Slack webhook."""
+        slack_message = self.get_data("slack", notifiable, notification)
+        recipients = self.get_recipients(notifiable)
+        mode = self.get_sending_mode(recipients)
+        slack_message = slack_message.mode(mode)
+
+        if mode == self.WEBHOOK_MODE:  # and not slack_message._webhook:
+            slack_message = slack_message.to(recipients)
+        elif mode == self.API_MODE:
+            # if not slack_message._channel:
```
josephmancuso, line 48: old code?

src/masonite/notification/drivers/SlackDriver.py
```diff
@@ -0,0 +1,117 @@
...
+        for recipient in recipients:
+            if recipient.startswith("https://hooks.slack.com"):
+                modes.append(self.WEBHOOK_MODE)
+            else:
+                modes.append(self.API_MODE)
+        if len(set(modes)) > 1:
+            raise NotificationException("Slack sending mode cannot be mixed.")
+        return modes[0]
+
+    def send_via_webhook(self, slack_message):
```
josephmancuso, line 71: Whats the benefit of having both sending via webhook and sending via API?

src/masonite/notification/drivers/vonage/VonageDriver.py
```diff
@@ -0,0 +1,73 @@
...
+        client = vonage.Client(
+            key=self.options.get("key"), secret=self.options.get("secret")
+        )
+        return Sms(client)
+
+    def send(self, notifiable, notification):
+        """Used to send the SMS."""
+        sms = self.build(notifiable, notification)
+        client = self.get_sms_client()
+        # TODO: here if multiple recipients are defined in Sms it won't work ? check with Vonage API
```
josephmancuso, line 42: loop here and send to each one?

src/masonite/notification/drivers/vonage/VonageDriver.py
```diff
@@ -0,0 +1,73 @@
...
+        response = client.send_message(sms)
+        self._handle_errors(response)
+        return response
+
+    def queue(self, notifiable, notification):
```
josephmancuso, line 47: remove for now

tests/integrations/config/notification.py
```diff
@@ -0,0 +1,21 @@
...
+        "secret": os.getenv("VONAGE_SECRET", ""),
+        "sms_from": os.getenv("VONAGE_SMS_FROM", "+33000000000"),
+    },
+    "database": {
+        "connection": "sqlite",
+        "table": "notifications",
+    },
+}
+
+DRY = False
```
josephmancuso, line 21: Whats this thing?

josephmancuso (changes requested):
Did another complete code review. Also manually tested this feature and with some changes i already did i think it works really well. Before i do more manual testing i just did another code review where we need to talk about a few things i commented on.

Also need to remove the queuing stuff. I think we'll have to revisit that in the future

There is also some TODO we need to address.

Once those changes are made i can retest some of the other features like the database notifications and broadcast drivers

girardinsamuel:
I will take a look and make changes beginning of next week !

src/masonite/notification/Notifiable.py
```diff
@@ -0,0 +1,63 @@
...
+    """Notifiable mixin allowing to send notification to a model. It's often used with the
+    User model.
+
+    Usage:
+        user.notify(WelcomeNotification())
+    """
+
+    def notify(self, notification, drivers=[], dry=False, fail_silently=False):
+        """Send the given notification."""
+        from wsgi import application
```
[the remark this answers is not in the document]
girardinsamuel: Why not ? 
It's the way used in a lot of different places in M3/4 isn't it ?

src/masonite/stubs/notification/create_notifications_table.py
girardinsamuel: It's now fixed, I just need to update the database in the sqlite file.

girardinsamuel: I actually used a big_increments, no need to hide the id I guess

src/masonite/notification/drivers/SlackDriver.py
girardinsamuel: Based on our Slack discussions. Both modes are useful (have pros and cons). Maybe later we could deprecate one or another ...

tests/integrations/config/notification.py
girardinsamuel: This is to globally disable notifications sending. I have found it useful when running commands in production for example. This is just a global switch controlled by a setting / environment variable

src/masonite/notification/Notifiable.py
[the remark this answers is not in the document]
josephmancuso: Yeah this is fine now that I've thought about i

josephmancuso (approved):

girardinsamuel:
Yeaaahhhhh 🚀

PR #54, opened 2021-02-23
Feature/7

### LINKED ISSUE #7: Add Testcases to quickly test scenarios
Database support should be separate. Possibly even connected to the Masonite ORM. You should be inheriting a class on your test case like:

```python
class TestUsers(TestCase, DatabaseTransactions):
    # ...
```

josephmancuso:
Closes #7


src/masonite/foundation/Kernel.py
```diff
@@ -91,6 +93,7 @@ def register_framework(self):
             "sign", Sign("-RkDOqXojJIlsF_I8wWiUq_KRZ0PtGWTOZ676u5HtLg=")
         )
         self.application.bind("base_url", "http://localhost:8000")
+        self.application.bind("resolver", DB)
```
girardinsamuel, line 96: @josephmancuso Why is it called `resolver` ? Is it bound only for being used in tests, or do you plan that we can resolve this in other parts of the app ?

PR #56, opened 2021-02-25
added hashids

josephmancuso:
Adds all the features of the masonite essentials package so we can archive that on release


josephmancuso (commented):
reviewed

PR #57, opened 2021-02-25
added event listener for exceptions

### LINKED ISSUE #37: Add exception listeners
This is a feature of Masonite where we can listen to multiple exceptions. But maybe this should tie into the Masonite events system ..

josephmancuso:
Closes #37 

@girardinsamuel here you can see how powerful this event system can be. Previously we did something similiar but this event system allows us to listen to all exceptions of the framework now. We can really easily add something like sentry to the app and listen to either a specific error: `masonite.exceptions.ZeroDivisionError` or we can listen to all exceptions: `masonite.exceptions.*` and send them all the Sentry or something. The sentry part if not in this PR this just adds support for something like that.

Or even for Masonite Logging

### DIFF
--- a/src/masonite/exceptions/ExceptionHandler.py
+++ b/src/masonite/exceptions/ExceptionHandler.py
@@ -29,6 +29,9 @@ class ExceptionHandler:
     def handle(self, exception):
         response = self.application.make("response")
         request = self.application.make("request")
+        self.application.make("event").fire(
+            f"masonite.exception.{exception.__class__.__name__}", exception
+        )
         handler = Handler(exception)
         handler.integrate(StackOverflowIntegration())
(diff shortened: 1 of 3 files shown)


josephmancuso (commented):
reviewed

src/masonite/exceptions/ExceptionHandler.py
```diff
@@ -29,6 +29,9 @@ def get_config_options(self, driver=None):
     def handle(self, exception):
         response = self.application.make("response")
         request = self.application.make("request")
+        self.application.make("event").fire(
```
girardinsamuel, line 32: @girardinsamuel 
So here it means that every exception caught by the handler is fired as an event with the name of the exception ? nice 💪

PR #62, opened 2021-02-26
Feature/20

### LINKED ISSUE #20: Add Websocket Broadcasting

josephmancuso:
Closes #20 

This PR adds websocket broadcasting


josephmancuso:
Still need to add private channel support in a way that will work with other drivers like Ably and Pubnub

josephmancuso:
## Usage

Usage is pretty simple.

First you can just fire normal events:

```python
def show(self, broadcast: Broadcast):
    broadcast.channel("channel-name", "event-name", {"key": "value"})
```

You can also emit an event name which is just a class that inherits `CanBroadcast`. Instead of a string as the event you can pass a class:

```python
class OrderProcessed(CanBroadcast):
    def __init__(self):
        self.order_id = 1

def show(self, broadcast: Broadcast):
    broadcast.channel("channel-name", OrderProcessed()})
```

**This will emit all of the properties of the class as the values**:

```python
{"order_id": 1}
```

josephmancuso (commented):
reviewed

josephmancuso:
@girardinsamuel Not 100% sure if I'm going to add the Ably and Pubnub drivers into here. I don't want to waste too much time on a feature I haven't seen a lot of people even use in Masonite ... 🤔 Maybe we just wait for the demand before adding other drivers to this

josephmancuso:
@girardinsamuel i'm going to merge this in for now but if you have spare time and can do a code review of this that would be great

PR #66, opened 2021-03-04
Feature/45 - Masonite API

### LINKED ISSUE #45: Add Masonite API back to core
Need to look at the Masonite API package and merge that into this one as we did with Masonite-scheduler.

This will need to be spec'd out though in order to make sure we hit all the core requirements when we do the back port.

**Full feature requirements to come**

josephmancuso:
Closes #45

- [x] Authentication classes (for getting tokens)
- [x] API Guard
- [x] Authentication Middleware to verify token
- [ ]  Resource classes
- [ ] Authentication Drivers (`token`)
- [ ] Hashing tokens


josephmancuso:
Theres only one failing test here .. But its a huge problem .. A problem that could mean a rewrite of the routing engine ..

The problem is that we have 2 different middleware stacks .. a `web` stack and a `api` stack. The `web` stack has things like cookie encryption, CSRF protection, etc.

The `api` stack has things like token authentication.

One way to accomplish this is have 2 route groups. Route group 1 has the necessary web routes. That group can then have a `web` middleware on it that runs those web routes. The second route group has all your API routes. It then has the api middleware stack on it.

This actually works well when running the server but for some reason the tests don't work ..

josephmancuso:
I fixed this issue using `setRoutes` in the test case. Not sure if this is a great idea as it's going to create a lot of confusion i think if someone runs into this

src/masonite/api/resources/Resource.py
```diff
@@ -0,0 +1,29 @@
+from ...request import Request
+
+
+class Resource:
+    def _show(self, request: Request, user_id):
```
girardinsamuel, line 5: I guess here we will need to specify the model as a class attribute or a method ?
```python
class UserResource(Resource):
     model = User
     # or 
     def get_model(self):
         from my.specific.path import User
         return User
```

src/masonite/foundation/HttpKernel.py
```diff
@@ -10,13 +10,24 @@
...
+from ..utils.structures import load
 
 
 class HttpKernel:
 
-    http_middleware = []
-    route_middleware = {"web": [EncryptCookies, SessionMiddleware, VerifyCsrfToken]}
+    http_middleware = [ResponseMiddleware]
+
+    route_middleware = {
```
girardinsamuel, line 21: What about using the API session based ? I am a bit lost with the guards versus middleware roles in here 🤔 
But as I was saying in the #45 issue, I guess it would be cool to be able to use the API and authenticate requests with:
- session (stored in cookie)
- JWT tokens

Maybe we are not naming things the same way but here I have the feeling that it's not currently possible with this SessionMiddleware only in "web" route stack

src/masonite/headers/HeaderBag.py
```diff
@@ -35,6 +35,8 @@ def load(self, environ):
         for key, value in environ.items():
             if key.startswith("HTTP_"):
                 self.add(Header(key, value))
+            elif key.lower().replace("-", "_") in ("content_type", "content_length"):
```
girardinsamuel, line 38: Why is the use case for this if ? When is it used ?

src/masonite/stubs/api/Resource.py
```diff
@@ -0,0 +1,14 @@
+from masonite.api.resources import Resource
+
+
+class __class__(Resource):
+    def show(self, request, __record___id):
```
girardinsamuel, line 5: should we add a line (that the developer will complete after scaffolding ?) ?
```python
__model__ =
```

src/masonite/utils/helpers.py
```diff
@@ -520,6 +524,20 @@ def info(self, message):
...
+    """Generate a random string based on the length given.
+
+    Keyword Arguments:
+        length {int} -- The amount of the characters to generate (default: {4})
+
+    Returns:
+        string
+    """
+    return "".join(
+        random.choice(string.ascii_uppercase + string.digits) for _ in range(length)
```
girardinsamuel, line 537: Should we use a constant time generation method for better security here ? Or is it not useful ?

src/masonite/stubs/api/Resource.py
girardinsamuel: ah no okay I got it ! The model is written in all methods, I guess it's more flexible ?
I was thinking that we would do something like this instead : 
```python
class __class__(Resource):
     __model__ = User

    def show(self, request, __record___id):
         return self.__model__.find(__record___id)

    def index(self, request):
        return self.__model__.paginate(request.input("limit", 20), request.input("page", 1))
```

tests/integrations/controllers/api/UserResource.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class UserResource(Resource):
+    def show(self, request, user_id):
+        return User.find(user_id)
+
+    def index(self, request):
+        return User.paginate(request.input("limit", 10), request.input("page", 1))
+
+    def to_dict(self, request, model):
```
girardinsamuel, line 12: When seeing this I guess it covers lots of use cases but it would be also cool to be able to have a "quickest alternative" where we would specify fields only as in Django Rest Framework.
Something like this:
```
# ...
     return ["name", "email"]

or __fields__ = ["name", "email"]
```
a bit like visible and hidden attrs on ORM models definition

tests/integrations/web.py
```diff
@@ -1,14 +1,26 @@
...
+    Route.get("/test", "WelcomeController@test"),
+    Route.get("/emit", "WelcomeController@emit"),
+    Route.get("/view", "WelcomeController@view"),
+    Route.get("/mail", "MailableController@view"),
+    middleware=("web",),
+)
+
+Route.group(
+    Route.get("/protected", "WelcomeController@protect"),
+    Route.get("/users", "api.UserResource@_index").name("users.index"),
```
girardinsamuel, line 19: Can we register all resource endpoints at once too ?  something like:
```
     Route.get("/users", "api.UserResource").name("users.")
```

src/masonite/api/guards/APIGuard.py
```diff
@@ -0,0 +1,77 @@
...
+        attempt = self.options.get("model")().attempt_by_id(user_id)
+
+        if attempt and not self.options.get("once"):
+            self.application.make("request").cookie("token", attempt.remember_token)
+            return attempt
+
+        return False
+
+    def reset_password(self, username, new_password):
+        """Login a user by the user ID.
```
girardinsamuel, line 55: docstring needs update 😉

girardinsamuel (commented):
Nice work so far ! 
Just a bit confused with different auth capabilities (but I guess you will be able to give me some clarifications 😉).
I really like how it's done 🚀

josephmancuso:
We can revisit this now that the routing is fixed. The routing is what caused this PR to grind to a halt because I wasn't sure how to register the routes properly

josephmancuso:
Now that all the core features are merged in and being tested we can try revisiting this. We'll need to start from scratch pretty much though because theres so many conflicts in this branch.

I don't think i even got far enough to have anything here be viable so that should be fine

PR #68, opened 2021-03-15
Fix password resets table migration

girardinsamuel:
Here the password reset table name was incorrect.
But when edited the db I lost the password reset table data so some tests are not working. We could add a seed to create a password reset entry as I have done in #69 for the user table.


josephmancuso (approved):

josephmancuso (approved):

josephmancuso:
I fixed this directly on the master. I think this PR is failing because it gets rid of the SQLite database we use for testing

PR #78, opened 2021-03-23
Feature: add way to override test response

girardinsamuel:
I am thinking about packages and futures extensions of the framework: if an external package wants to bring new testing assertions how would be able to do it ?

Final chosen approach, can extend with multiple classes from different packages !

```python
# in package provider of package 1
self.application.make("tests.response").add(Package1TestResponse)
# in package provider of package 2
self.application.make("tests.response").add(Package2TestResponse)

# in the test

class ProjectTests(TestCase):

    def my_test(self):
        # can use default assertions and custom from different packages
        self.get("/").assertContains("welcome").assertPackage1Stuff().assertPackage2Stuff()

### CODE REVIEW

**Comment by josephmancuso** [comment]
I don't this will work right? This works for a single package but if i install 2 packages, both with testing assertions, this doesn't really work right?

* Package A binds `tests.response` to `PackageAResponse` that has 2 assertion methods: A,B
* Package B binds `tests.response` to `PackageBResponse` that has 2 assertion methods: C,D 

but now you lose the methods for A and B

**Comment by girardinsamuel** [comment]
@josephmancuso Hmm you're right ... !

I tried a new solution to dynamically build the test response classes.
```python
# in package provider of package 1
self.application.make("tests.response").add(Package1TestResponse)
# in package provider of package 2
self.application.make("tests.response").add(Package2TestResponse)

# in the test

class ProjectTests(TestCase):

    def my_test(self):
        # can use default assertions and custom from different packages
        self.get("/").assertContains("welcome").assertPackage1Stuff().assertPackage2Stuff()

```

josephmancuso (approved):
Seems really simple actually. This looks different from the last time i checked this PR. I like this a lot

PR #86, opened 2021-04-09
WIP: add request validation feature

girardinsamuel:
For now I am playing with different ideas...

I like how it's done in Laravel, two approaches: manual validation or automatic with redirect handling and flashing into session.

## Specifications
TODO


josephmancuso:
I think this is a phase 2 feature here

girardinsamuel:
> I think this is a phase 2 feature here

I agree, and it should not be too hard to implement now 😉

PR #87, opened 2021-04-09
Add optional helper

girardinsamuel:
Add `optional` helper which can be useful to get attributes on model when we are not sure they exist.

```python
# if user has no profile yet, this won't raise an exception
optional(User.find(1).profile).avatar #== None

# default is configurable
optional(User.find(1).profile, default=False).is_verified #== False
```

It can work when using method but `is_method` should be set to `True` (I did not find a better way):
```python
optional(User.find(1).profile, is_method=True).create_avatar() #== None, won't raise an exception is user has no profile
```

## remarks
- Idea based on Laravel `optional` helper
- I often needed this to avoid doing some bad if statements
- It can be used with any python objects (not only ORM models), so can be useful in some business logic sometimes


Marlysson:
The trick using a internal class returning an object with metamethod was mindblow to me.

Awesome!!!

src/masonite/utils/helpers.py
```diff
@@ -539,3 +539,28 @@ def password(password_string):
...
+
+
+def optional(attribute, default=None, is_method=False):
+    """Wrap an object on which any attributes/methods can be called
+    without raising an error if it does not exist. It will return a default value of None instead,
+    which can be overriden.
+
+    If you want to call method on the wrapped object, is_method should be set to True."""
+
+    class OptionalWrapper(object):
```
josephmancuso, line 551: Classes in Python 3 don't need to inherit the object class. This can be removed. 

```suggestion
    class OptionalWrapper:
```

josephmancuso (changes requested):

josephmancuso:
This exists in M3. Is there a reason you rewrote this? I think the one in M3 takes into account the method thing. Maybe you can either port that one over or draw inspiration to how i did that

josephmancuso:
https://github.com/MasoniteFramework/masonite/blob/383cf4607700bfa442fa6bbf33c764035eee2fee/src/masonite/helpers/optional.py#L12

girardinsamuel:
> https://github.com/MasoniteFramework/masonite/blob/383cf4607700bfa442fa6bbf33c764035eee2fee/src/masonite/helpers/optional.py#L12

I did not check this was already available 😓 at least I have learned stuff on the way ! 😅 
So I will backport this instead.

girardinsamuel:
@josephmancuso I updated this helper with Masonite 3 code.
I just added a ability to provide a default value different from `None`

josephmancuso (approved):

PR #91, opened 2021-04-16
fixed routing issues

### LINKED ISSUE #90: Fix Routing
One of the issues with routing is that there is really no good encapsulation with the current structure.

Meaning that there are a lot of side effects, setting routing in 1 file could change routing in another file

josephmancuso:
Closes #90 

@girardinsamuel I took a deep look into this and I think we might be good. There is still an issue with having 2 parent route groups in the same file but not sure if we can around that ..

### DIFF
--- a/src/masonite/foundation/HttpKernel.py
+++ b/src/masonite/foundation/HttpKernel.py
@@ -38,13 +38,26 @@ class HttpKernel:
 
         self.application.bind(
             "router",
-            RouteCapsule(Route.group(*pydoc.locate(self.application.make("routes.web")).routes, middleware="web")),
+            RouteCapsule(
+                Route.group(
+                    *pydoc.locate(self.application.make("routes.web")).routes,
+                    middleware="web"
+                )
+            ),
         )
 
-        self.application.make('router').add(
-            Route.group(Route.get('/test', 'WelcomeController@api'), prefix="/api", middleware="api") 
+        self.application.make("router").add(
+            Route.group(
+                Route.get("/test", "WelcomeController@api"),
+                prefix="/api",
+                middleware="api",
+            )
         )
 
-        self.application.make('router').add(
-            Route.group(load_routes(self.application.make("routes.api")), middleware="api", prefix="/api")
+        self.application.make("router").add(
+            Route.group(
+                load_routes(self.application.make("routes.api")),
+                middleware="api",
+                prefix="/api",
+            )
         )
--- a/src/masonite/foundation/HttpKernel.py
+++ b/src/masonite/foundation/HttpKernel.py
@@ -11,6 +11,7 @@ from ..middleware import (
 )
 from ..routes import RouteCapsule, Route
 import pydoc
+from ..utils.structures import load_routes
 
 
 class HttpKernel:
@@ -37,5 +38,13 @@ class HttpKernel:
 
         self.application.bind(
             "router",
-            RouteCapsule(*pydoc.locate(self.application.make("routes.web")).routes),
+            RouteCapsule(Route.group(*pydoc.locate(self.application.make("routes.web")).routes, middleware="web")),
+        )
+
+        self.application.make('router').add(
+            Route.group(Route.get('/test', 'WelcomeController@api'), prefix="/api", middleware="api") 
+        )
+
+        self.application.make('router').add(
+            Route.group(load_routes(self.application.make("routes.api")), middleware="api", prefix="/api")
         )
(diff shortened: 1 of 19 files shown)


src/masonite/foundation/HttpKernel.py
```diff
@@ -37,5 +38,13 @@ def register_routes(self):
 
         self.application.bind(
             "router",
-            RouteCapsule(*pydoc.locate(self.application.make("routes.web")).routes),
+            RouteCapsule(Route.group(*pydoc.locate(self.application.make("routes.web")).routes, middleware="web")),
+        )
+
+        self.application.make('router').add(
```
girardinsamuel, line 44: This is just for testing right ?

josephmancuso:
## Current Way (M3)

```python
from masonite.routes import Get, Post

ROUTES = [
    Get('/url/1', 'WelcomeController@show'),
    Post('/url/1', 'WelcomeController@post'),
]
```

## New Way (M4)

```python
from masonite.routes import Route

Route.get('/url/1', 'WelcomeController@show')
Route.post('/url/1', 'WelcomeController@post')
```

## Proposal (Fixed M4 issues)

```python
ROUTES = [
   Route.get('/url/1', 'WelcomeControlle@show'),
   Route.post('/url/1', 'WelcomeControlle@post'),
   Route.group([
    
    ])
]
```

josephmancuso:
@girardinsamuel I think we can fix this without rewriting much by just removing the global state of the Route class and putting them still inside that global ROUTES list we had before in M3

```python
ROUTES = [
   Route.get('/url/1', 'WelcomeControlle@show'),
   Route.post('/url/1', 'WelcomeControlle@post'),
   Route.group([
    
    ])
]
```

girardinsamuel:
> @girardinsamuel I think we can fix this without rewriting much by just removing the global state of the Route class and putting them still inside that global ROUTES list we had before in M3

Looks like a good idea to mix the two approaches !

tests/integrations/web.py
```diff
@@ -2,13 +2,15 @@
...
+    Route.get("/", "WelcomeController@show").name("welcome"),
+    Route.post("/", "WelcomeController@show"),
+    Route.post("/upload", "WelcomeController@upload").name("upload"),
+    Route.get("/test", "WelcomeController@test"),
+    Route.get("/emit", "WelcomeController@emit"),
+    Route.get("/view", "WelcomeController@view"),
+    Route.get("/mail", "MailableController@view"),
+]
 
 Broadcast.routes()
```
girardinsamuel, line 15: With this new approach what about those:
```python
Broadcast.routes()
Auth.routes()
```
How to handle this ? Should we do somehting like : 
```python
ROUTES += Brodcast.routes()
ROUTES += Authroutes()
```

tests/tests/test_testcase.py
```diff
@@ -55,14 +55,14 @@ def test_fake_time_in_future(self):
         self.fakeTimeInFuture(1, "months")
         self.assertEqual(pendulum.now().diff(real_now).in_months(), 1)
 
-    def test_fake_time_in_past(self):
```
girardinsamuel, line 58: Did you had issues with this test 🤔 ??? It's working on master normally

tests/tests/test_transactions.py
```diff
@@ -13,7 +13,7 @@ def setUp(self):
     def test_can_use_transactions(self):
         User.create({"name": "john", "email": "john6", "password": "secret"})
 
-    def test_assert_deleted(self):
-        user = User.find(1)
-        user.delete()
-        self.assertDeleted(user)
+    # def test_assert_deleted(self):
```
girardinsamuel, line 16: With this one though there is an issue do you know how to solve it ?

girardinsamuel (commented):
I like this new approach 💪

tests/tests/test_testcase.py
josephmancuso: I had to remove some of these tests as they were throwing a weird SQLite database connection already closed issue. Which may be a bigger issue with the ORM I'm not sure. I dove into it but I didn't have enough time to complete the rabbit hole journey as my debugging I couldn't find the issue

tests/tests/test_transactions.py
josephmancuso: I looked into it but I could't find the issue. Will have to take a look at it

josephmancuso (commented):
reviewed

tests/tests/test_testcase.py
girardinsamuel: Okay let's dig into that later !

tests/integrations/web.py
girardinsamuel: @josephmancuso what about this ?

PR #93, opened 2021-04-16
Add database testing assertions

girardinsamuel:
This PR adds useful and simple database testing assertions:

```python
self.assertDatabaseCount("users", 3)
self.assertDatabaseHas("users", {"name": "Sam", "id": 2})
self.assertDatabaseMissing("users", {"name": "John"})

user = User.find(1)
user.delete()
self.assertDeleted(user)
self.assertSoftDeleted(user) # == same but for models with soft delete mixin
```

### DIFF
--- a/tests/TestCase.py
+++ b/tests/TestCase.py
@@ -144,3 +144,34 @@ class TestCase(TestCase):
         """Restore the service previously mocked to the original one."""
         original = self.original_class_mocks.get(binding)
         self.application.bind(binding, original)
+
+    def assertDatabaseCount(self, table, count):
+        self.assertEqual(self.application.make("builder").table(table).count(), count)
+
+    def assertDatabaseHas(self, table, query_dict):
+        self.assertGreaterEqual(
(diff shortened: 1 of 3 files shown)


Marlysson:
@girardinsamuel 

The methods **assertDatabaseHas** and **assertDatabaseMissing** is related to "exists" and "notExists" expressions? 
By using Has or Missing can be confusing with some other methods that handle attributes in some couple of data like assertHasJson ( has some attribute or couple of attributes) assertHasHeader ( has some attribute in header request/response)

I've seen in Laravel's documentation and it's same api that you defined. What do you think of changing the api to "assert[Database]"Exists" and "assert[Database]NotExists" to be more clear that mean exists or not some data based in some query?

girardinsamuel:
Yes, here the idea is that the table "has" some records satisfying the given query...

We have already has and missing words in existing M4 assertions, so I was thinking to stick with it. But I can made some changes if you think it would be more clear !

@josephmancuso what's your opinion on this ?

josephmancuso (approved):

PR #94, opened 2021-04-18
Add commands testing assertions

girardinsamuel:
Now we can run `self.craft("key")` in tests (or any command name with some arguments) and then be able to assert command outputs/errors:

```python
self.craft("key")
    .assertSuccess()
    .assertExactOutput("....")
    .assertOutputContains("...")
    .assertOutputMissing("....")
    .assertHasErrors()
    .assertExactErrors("....")
```

To pass arguments to command it's a simple string
```python
self.craft("new", "new_project --branch 2.3 --repo MasoniteFramework/secret")
```


josephmancuso (approved):
This is really cool!!

josephmancuso (approved):

PR #95, opened 2021-04-29
Fix using masonite TestCase externally

girardinsamuel:
For now when running unit tests based on masonite4 TestCase class, from a project or in an external package, some imports errors occurs from masonite4.

I found weird that TestCase wasn't imported from the package but from tests/ folder as we need a fully featured TestCase.

This PR is thus moving the TestCase class into src/masonite/tests and fixing imports to be relative.


josephmancuso:
The test case should be importing from the tests/TestCase.py file. This is so they can override some things like where the wsgi container is and some other values we throw in there

girardinsamuel:
I am not sure to understand what you mean ..

I have opened this PR because I get the following errors with the current masonite version when I try to create and run tests from a M4 project:
```
../../../../.pyenv/versions/masonite-admin/lib/python3.9/site-packages/masonite/tests/__init__.py:3: in <module>
    from .HttpTestResponse import HttpTestResponse
../../../../.pyenv/versions/masonite-admin/lib/python3.9/site-packages/masonite/tests/HttpTestResponse.py:2: in <module>
    from src.masonite.views import View
E   ModuleNotFoundError: No module named 'src.masonite.views'
```

PR #96, opened 2021-04-30
Fix TestCase imports

girardinsamuel:
Here is a new PR which fixes the imports without refactoring the TestCase location.


girardinsamuel:
The issue in tests is not related to this PR 🤔 (it's related to db testing assertions and I don't know how to fix it).

girardinsamuel:
Finally this is not what we need to do.

PR #97, opened 2021-04-30
Fix location of TestCase class

girardinsamuel:


girardinsamuel:
As discussed in Slack I moved the TestCase implementation in src/masonite/tests.

josephmancuso:
Yeah this is exactly right. Great Job!

PR #100, opened 2021-05-07
added static asset

### LINKED ISSUE #83: Add static file serving
I played with file uploads for `masonite-admin` 😉 . It's working great. But then I tried to display a user avatar (e.g to serve an image from masonite) it did not work => I did not found static files handling feature. It was there in M3.

I guess it have to be implemented, the same way ?

- static helpers for view
- static url path
- binding to multiple dirs
- interaction with storage ? For now disk().store(...) return the full path to the file. In database we want to store a relative path that the static helper will be able to resolve later I guess ?

josephmancuso:
Closes #83 

Theres 2 similar things here we need to separate:

* Static files (CSS, Logos, JS files, etc)
* Assets (Basically just user Uploads)

@girardinsamuel This code is mainly pulled from Masonite 3.

## Assets

You can set the paths on the disk. These paths will be the base directories where you assets are located:

```python
DISKS = {
    "default": "s3",
    "s3": {
        "driver": "file",
        "path": "https://s2.west2.amazon.com/public"
        #
    },
}
```

You can then use a new `asset()` view helper to concatenate the paths:

```html
<img src="{{ asset('s3', 'relative/path')"> <!-- https://s2.west2.amazon.com/public/relative/path -->
```

You can also structure the path as a dictionary and use dot notation:

```python
DISKS = {
    "default": "s3",
    "s3": {
        "driver": "file",
        "path": {
             'avatars': "https://s2.west2.amazon.com/public"
        }
        #
    },
}
```

```html
<img src="{{ asset('s3.avatars', 'relative/path')"> <!-- https://s2.west2.amazon.com/public/relative/path -->
```

## Static Files

Static files are treated a bit differently. These are files like CSS and JS files that do not change or get added throughout the life of the application deployment.

The way to do this is already really built in. We just alias the directory we want to serve static files from.

Most directories by default are aliased with the `/static/` alias. So you can alias any number of directories to `/static`. This configuration:

```python
storage.add_storage_assets(
            {
                # folder                             # template alias
                "tests/integrations/storage/static": "static/",
                "tests/integrations/storage/compiled": "static/",
                "tests/integrations/storage/uploads": "static/",
                "tests/integrations/storage/public": "/",
            }
        )
```

Allows any files in any of those directories on the left to be aliased in a template to `/static`:

```html
<link rel="stylesheet" href="/static/main.css">
```

I don't think we need a helper for this as it's already sorta aliased.

@girardinsamuel what are your thoughts?

### DIFF
--- a/tests/integrations/templates/welcome.html
+++ b/tests/integrations/templates/welcome.html
@@ -8,9 +8,14 @@ Hello. Setup chat connection here
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Document</title>
     <script src="https://js.pusher.com/7.0/pusher.min.js"></script>
+    <link rel="stylesheet" href="/static/main.css">
 </head>
 <body>
     <h1>Welcome</h1>
+
+    <div>
+        <img src="{{ asset('local', 'original/image-20150902-6700-t2axrz.jpg') }}"
+    </div>
 
     <script>
         const pusher = new Pusher("", {
--- a/tests/integrations/templates/welcome.html
+++ b/tests/integrations/templates/welcome.html
@@ -13,10 +13,6 @@ Hello. Setup chat connection here
 <body>
     <h1>Welcome</h1>
 
-    <div>
-        <img src="{{ asset('local', 'original/image-20150902-6700-t2axrz.jpg') }}"
-    </div>
-
     <script>
         const pusher = new Pusher("", {
             authEndpoint: '/broadcasting/authorize',
--- a/src/masonite/providers/HelpersProvider.py
+++ b/src/masonite/providers/HelpersProvider.py
@@ -1,5 +1,6 @@
 import builtins
 from ..providers import Provider
+from ..utils.helpers import AssetHelper
 
 
 class HelpersProvider(Provider):
@@ -18,6 +19,7 @@ class HelpersProvider(Provider):
                 "auth": request.user,
                 "route": self.application.make("router").route,
                 "cookie": request.cookie,
+                "asset": AssetHelper(self.application).asset,
                 "url": lambda name, params={}: (
                     self.application.make("base_url")
                     + self.application.make("router").route(name, params)
--- a/src/masonite/providers/HelpersProvider.py
+++ b/src/masonite/providers/HelpersProvider.py
@@ -1,6 +1,6 @@
 import builtins
 from ..providers import Provider
-from ..utils.helpers import AssetHelper
+from ..utils.helpers import AssetHelper, UrlHelper
 
 
 class HelpersProvider(Provider):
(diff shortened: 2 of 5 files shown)


tests/integrations/templates/welcome.html
```diff
@@ -8,10 +8,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Document</title>
     <script src="https://js.pusher.com/7.0/pusher.min.js"></script>
+    <link rel="stylesheet" href="/static/main.css">
 </head>
 <body>
     <h1>Welcome</h1>
 
+    <div>
+        <img src="{{ asset('local', 'original/image-20150902-6700-t2axrz.jpg') }}"
```
girardinsamuel, line 17: maybe put a more generic image here ?

girardinsamuel (commented):
It's what I had in mind.

But there is still the question of how to display the user uploaded file ? Using the helper or not, the path should be in what is served by whitenoise ?
Do you agree ?

girardinsamuel:
Also maybe it would be useful to add a `url()` helper in the same fashion ? But this one would get the full url to an asset ?
> Use case e.g. would be sending a link to download an invoice or a specific asset in an email

```python
    def url(self, alias, file_name):
        disks = load(self.app.make("config.filesystem")).DISKS
        base_url = self.app.make("base_url").rstrip("/")  # just ensure that no slash is appended to the url
        if "." in alias:
            alias = alias.split(".")
            location = disks[alias[0]]["path"][alias[1]]
            if location.endswith("/"):
                location = location[:-1]

            return "{base_url}/{}/{}".format(location, file_name)

        location = disks[alias]["path"]
        if isinstance(location, dict):
            location = list(location.values())[0]
            if location.endswith("/"):
                location = location[:-1]

        return "{base_url}/{}/{}".format(location, file_name)
```

tests/integrations/templates/welcome.html
josephmancuso: This is just an example, it will be removed on final commit

josephmancuso:
@girardinsamuel yeah we can do this off of the base_url

josephmancuso:
@girardinsamuel added url helper. There was also a url helper but thats really the route helper .. so i just changed their names

josephmancuso:
@girardinsamuel so the way you would do this for an upload is you would:

* set the base path in the disk
* Upload the asset and get the relative url
* Use the `asset()` helper to help append the absolute path to the relative path:

```html
<img src="{{ asset('s3', user.avatar)">
```

josephmancuso:
i'll merge it in now but we can talk about it more if you want

girardinsamuel:
> i'll merge it in now but we can talk about it more if you want

Yes I would like i still don't manage to show an uploaded file in a view with M4...We can discuss it on slack

PR #102, opened 2021-05-10
Add driver() option to Mailable

girardinsamuel:
- Refactored mail drivers into `mail` feature folder
- Added driver() method to Mailable

Now when sending a mailable, if no driver is specified in send() the mailable driver will be used and if not set it will fallback to the default driver in config: 
```python
class Welcome(Mailable):

    def build(self):
        return (
            #...
            .driver("mailgun")
        )

class Welcome2(Mailable):

    def build(self):
        return (
            #...
        )

app.make("mail").mailable(Welcome()).send(driver="terminal") # send with terminal
app.make("mail").mailable(Welcome()).send() # send with mailgun
app.make("mail").mailable(Welcome2()).send() # send with default driver in config/mail
```


josephmancuso (approved):
Love it

PR #103, opened 2021-05-10
WIP: facades implementation experiment

girardinsamuel:
@josephmancuso From Slack discussion I played a bit with Facade implementation in M4.
 - Seems to work in tests
 - But in IDE, classes are not resolved (it does not prevent code to work, but that's not very satisfying). As Python is not typed, I guess it's not easily fixable ..


josephmancuso:
I don't love the implementation. I don't like the globals kinda of thing. I don't remember i said we should go this way or not .. it sounds familiar lol. So much going on lately. 

Globals in Python is annoying because text editors yell and its confusing on where the heck this call is coming from. Most people look to the top of the file so they know what class it is and where it is but when its using a builtin .. its hard to track down whats going on

josephmancuso:
Added an experimental suggestion, let me know your thoughts.

To create a facade you should have to do something like this:

```python
from masonite.facades import Facade
class Response(metaclass=Facade):
    key = 'response' # The key to make from the container
```

Then you can start using this everywhere. Like in a controller:

```python
from masonite.facades import Response

def show(self):
    Response.cookie('key', 'value')
```

girardinsamuel:
> ```python
> from masonite.facades import Response
> 
> def show(self):
>     Response.cookie('key', 'value')
> ```

I like this implementation !
But I am not sure about the import magics ...
- in the tests you're using some facades without any imports
- in the exemple you posted above, you're importing the `Response` facade from `facades` module

I don't fully understand how it works.

josephmancuso (approved):

josephmancuso:
@girardinsamuel 

> in the tests you're using some facades without any imports

Those weren't my test they were yours. They were passing because i never removed your code so however it was working before lol

> in the exemple you posted above, you're importing the Response facade from facades module, but no facades are declared here...

Yeah you have to create a facade like any normal class

>  I don't understand how the facades are available anywhere without imports : I suppose it's because every registered facades are set into the builtins module with the line setattr(builtins, facade_name, self.application.make(container_name))

They are not. They should be imported and used manually like any normal Python class

> I understand that just calling the Response facade is resolving the response instance bound to the container

Yup. If you actually notice, the Facade classes are exactly the same as the model classes. That metaclass= is the same an ORM model.

PR #106, opened 2021-05-28
Add hashing feature

### LINKED ISSUE #101: Support Argon2 encryption method along with bcrypt

girardinsamuel:
Fixes #101.

Implements `hashing` feature - accessible with a facade - providing two hash protocols: `bcrypt` and `argon2`.
- bcrypt is the default
- each of them have sensible defaults (values from Laravel)
- we can make a hash, verify a hash and check if a hash needs to be hashed again (if parameters have been changed)
- argon2 needs `argon2-cffi` lib to be installed (optional)

This is used for now for hashing password before saving them in db, but can be used for hashing anything... 😅
New hash protocols can easily be added.


josephmancuso (changes requested):
This should look more like the other features i think

girardinsamuel:
> This should look more like the other features i think

I have started to rewrite this as proper feature, you're right it's better !

src/masonite/authentication/models/authenticates.py
```diff
@@ -18,19 +17,15 @@ def attempt(self, username, password):
         record_password = getattr(record, self.get_password_column())
         if not isinstance(record_password, bytes):
             record_password = bytes(record_password or "", "utf-8")
-        if bcrypt.checkpw(bytes(password, "utf-8"), record_password):
+        if Hash.check(password, record_password):
```
josephmancuso, line 20: Yup this is exactly how the facades should be used. We no longer have to import the wsgi file

src/masonite/facades/Hash.py
```diff
@@ -0,0 +1,5 @@
+from .Facade import Facade
+
+
+class Hash(metaclass=Facade):
+    key = "hash"
```
josephmancuso, line 5: perfect 👌

src/masonite/hashing/drivers/BcryptHasher.py
```diff
@@ -0,0 +1,25 @@
...
+        rounds = self.options.get("rounds")
+        salt = bcrypt.gensalt(rounds=rounds)
+        return str(bcrypt.hashpw(bytes(string, "utf-8"), salt))
+
+    def check(self, plain_string, hashed_string):
+        rounds = self.options.get("rounds")
+        # TODO: here we should take care of comparing use same rounds count
+        # byt bcrypt does not expose public api to do this !
+        # salt = bcrypt.gensalt(rounds=rounds)
+        # ref_hashed = bcrypt.hashpw(bytes(plain_string, "utf-8"), salt)
```
josephmancuso, line 22: Isn't this commented out code the correct way to do this?

josephmancuso, line 22: I am pretty sure with bcrypt, you do not need to specify the salt when checking the password. Just check the byte version of the supplied password with the hashed version of it

josephmancuso: The salt is already included in the password hash that we store in the database

josephmancuso (changes requested):

src/masonite/hashing/drivers/BcryptHasher.py
girardinsamuel: That's right.

girardinsamuel:
@josephmancuso feature ready ! Failing tests are the one from master (session, csrf stuff)

src/masonite/hashing/drivers/BcryptHasher.py
```diff
@@ -0,0 +1,28 @@
...
+        if not isinstance(hashed_string, bytes):
+            hashed_string = bytes(hashed_string or "", "utf-8")
+        return bcrypt.checkpw(bytes(plain_string, "utf-8"), hashed_string)
+
+    def needs_rehash(self, hashed_string):
+        # Bcrypt hashes have the format $2b${rounds}${salt}{checksum}. rounds is encoded as
+        # 2 zero-padded decimal digits. The prefix (2b) is never modified in make() function so we
+        # can assume that rounds value used when generating the hash is located at [4:6] indexes
+        # of the hash.
+        old_rounds_value = int(hashed_string[4:6])
```
josephmancuso, line 27: clever. I like that

josephmancuso (approved):

PR #108, opened 2021-05-29
General Improvements

josephmancuso:


src/masonite/container/container.py
```diff
@@ -211,21 +211,16 @@ def resolve(self, obj, *resolving_arguments):
...
-                    signature = "{}.{}.{}".format(
-                        obj.__module__, obj.__self__.__class__.__name__, obj.__name__
-                    )
-                    self._remembered[signature] = objects
-            return obj(*objects)
-        except (TypeError,) as e:
-            exception = ContainerError
-            exception.from_obj = obj
 
-            raise exception(str(e) + " while calling {}".format(obj)) from e
```
josephmancuso, line 228: @girardinsamuel removing this improves the stack trace. We were catching specific exceptions and then reraising a new exception. So anytime you application actually threw a typeError, you would never know where it came from because we were re-raising a new error. 

The weird part is I'm not sure why I ever did this. I don't see a need to do this ...

girardinsamuel: That's a nice fix because it was sometimes complex to find out what and where was the error raised !

girardinsamuel (approved):

josephmancuso:
I made some changes to the CSRF tokens to get it working through manual testing. Still need to work on some things as the encryption middleware was screwing everything up. Still need to look into it

tests/features/cache/test_file_cache.py
```diff
@@ -31,7 +31,7 @@ def test_will_not_get_expired(self):
         time.sleep(2)
         self.assertEqual(self.driver.get("expire"), None)
 
-    def test_will_get_not_yet_expired(self):
+    def test_will_get_not_get_expired(self):
```
girardinsamuel, line 34: Shouldn't it 
```suggestion
    def test_will_get_not_expired(self):
```

src/masonite/routes/HTTPRoute.py
```diff
@@ -269,9 +269,11 @@ def compile_route_to_regex(self):
         return regex
 
     def extract_parameters(self, path):
+        if not self.url_list:
```
girardinsamuel, line 272: when the route has no parameters this function was raising an error because no match found and .groups() was applied on NoneType

girardinsamuel (approved):

src/masonite/routes/HTTPRoute.py
josephmancuso: good catch

josephmancuso:
Accidentally committed the encryption cookie stuff to this branch

josephmancuso:
We now have to revisit the encrypt cookies middleware 😞 . This annoying middleware keeps breaking

josephmancuso (commented):
reviewed

PR #109, opened 2021-06-12
Add cookies to response

girardinsamuel:
For now cookies to send in the response are set by adding them on the request object => it feels a bit wrong.

This PR allows to set cookies to be returned in the response:

1/
```python
response.cookie("my_cookie", "value")
# can add options
response.cookie("my_cookie", "value", secure=True, http_only=False)
```

For now I have kept the request cookies rendering in the response, I guess it should be changed but I don't know how exactly.

2/ Add a fix in VerifyCSRFToken


src/masonite/middleware/route/VerifyCsrfToken.py
```diff
@@ -65,7 +65,7 @@ def in_exempt(self, request):
 
     def get_token(self, request):
         return (
-            request.header("X-CSRF-TOKEN")
-            or request.header("X-XSRF-TOKEN")
+            request.header("X-CSRF-TOKEN").value
```
girardinsamuel, line 68: Here I added a fix because headers are returned not the value... (we could also edit the request.header() method to return directly the value, I don't know what's best, maybe sometimes we would like to interact with the header object directly ?)

girardinsamuel: The problem now is that when a header is not request.header("MYHEADER") will return "" and `.value` won't work, what would you do here ?

girardinsamuel: I finally updated it the same way as for cookie so we don't have to call value

Marlysson: > The problem now is that when a header is not request.header("MYHEADER") will return "" and .value won't work, what would you do here ?

Can the `optional()` helper be used here? to case of don't throw a invalid property of None or empty value..

girardinsamuel: It could be actually it's the design that was not good I guess, using optional would feel like a patch.
I have finally use same method as for cookies (cf above comment)

src/masonite/foundation/response_handler.py
```diff
@@ -40,7 +40,9 @@ def response_handler(environ, start_response):
     print("response code", request.get_path(), response.get_status_code())
     start_response(
         response.get_status_code(),
-        response.get_headers() + request.cookie_jar.render_response(),
+        response.get_headers()
+        + request.cookie_jar.render_response()
```
josephmancuso, line 44: I'm not actually sure we even need this request cookies at all ..

josephmancuso (approved):
I think this PR is good. See the changes I did and if they make sense.

On every request, a browser should be sending the cookie in the request so we should always be able to GET cookies from the request class and SET cookies with the response class

josephmancuso:
The broken tests are unrelated so i will merge and fix them later

PR #112, opened 2021-06-28
WIP: add parsing of array or dict like form inputs

### LINKED ISSUE #104: Fix inputs when using array based inputs
In html we can do something like:

```html
<input type="text" name="comments[]" value="1">
<input type="text" name="comments[]" value="2">
```

When it submits it is currently submitting as 1 value:

```python
{
   "comments": 2
}
```

It should be submitting like:

```python
{
   "comments": [1,2]
}
```

When you submit an input like this:

```html
<input type="text" name="comments[description]" value="1">
<input type="text" name="comments[name]" value="2">
```

It should look like this:

```python
{
    "comments": {
        "description": 1,
        "name": 2
    }
}
```

and lastly it could get more advanced:

```html
<input type="text" name="comments[options][description]" value="1">
<input type="text" name="comments[options][name]" value="2">
```

should compile down to:

```python
{
    "comments": {
        "options": {
            "description": 1,
            "name": 2
        }
    }
}
```

Currently this advanced input doesn't work

girardinsamuel:
Fixes #104.

### DIFF
--- a/src/masonite/input/InputBag.py
+++ b/src/masonite/input/InputBag.py
@@ -1,6 +1,8 @@
 from .Input import Input
 from urllib.parse import parse_qs
-import email.parser
+from collections import defaultdict
+import urllib
+import re
 import json
 import cgi
 import re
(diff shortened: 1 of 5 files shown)


josephmancuso (commented):
I think this is fine? I'm pretty sure we can make this simpler though ... There is a urlib function that will parse this for us. I think we just need to do something like "If an input is a list with multiple elements, return a list, if its a list with a single value then save the value"

girardinsamuel:
> I think this is fine? I'm pretty sure we can make this simpler though ... There is a urlib function that will parse this for us. I think we just need to do something like "If an input is a list with multiple elements, return a list, if its a list with a single value then save the value"

I am already using this urllib function. But where it's not simple is for use case like this (nested dict) `comments[a][b]`

josephmancuso:
hmm I feel like it would have handled that ..

josephmancuso:
@girardinsamuel I fixed the rendering of all the single layer deep inputs i put in the ticket. The next part really is figuring out the multi layer ones like `user[options][user][name]`.

We're going to have to use some kind of recursion for this

josephmancuso:
@girardinsamuel I was able to fix 90% of the use cases. i could not solve the nested `user[options][name]` scenario. 

It's not super critical for that but we may have to revisit it sometime later

josephmancuso:
This solution is also a lot messier than i'd like but it works and this is difficult stuff lol

PR #115, opened 2021-06-29
Improving access to UrlHelper

### LINKED ISSUE #114: Improve access to url helper
I would like to use `UrlHelper(self.application).url()` helper outside of a view, it would be good to be able to do:
```python
from masonite.helpers..... import url

url(..)
```
instead of writing `UrlHelper(self.application).url()`

how should we do this ? Facade ? or just:

```python

# somewhere in helpers module

def url(path):
    from wsgi import application
    return UrlHelper(application).url(path)
```

girardinsamuel:
Fixes #114


tests/core/utils/test_helpers.py
```diff
@@ -21,3 +22,6 @@ def test_optional(self):
 
         self.assertEqual(optional(user, default=0).non_existing_attr, 0)
         self.assertEqual(optional(user, default=0).non_existing_method(), 0)
+
+    def test_url(self):
+        self.assertEqual(url.url("upload"), "http://localhost:8000/upload")
```
girardinsamuel, line 27: Here I need to do that ...
We would like to be able to do `url("upload")`...

It's because `url` is the instance bound to the container of UrlHelper and then we have to call the method...

girardinsamuel: Or maybe what we should bind is the url() method of UrlHelper directly ... ?

josephmancuso: That works too I think lol

girardinsamuel: It's working in the shell but does not seem to work in the unit test :/

girardinsamuel:
Ready to be merged it's working now ! 👌

josephmancuso (approved):

PR #116, opened 2021-06-30
Feature/63 - Request Validation

josephmancuso:
@girardinsamuel Just added a simple integration for validation here. 

I think i saw what you tried to do in #86 .. You tried to do how Laravel has the RequestValidators i think. Where the vaidation would throw an exception and we can catch it and handle the response.

This is really just the implementation we had in version 3 except a little updated for Masonite 4.

~It still uses masonite-validation since that package is still a good package.~

We have brought masonite-validation into core

## TODO:

- [x] Get scaffold commands to work properly
- [ ] Get templates to render error messages


josephmancuso:
We need to figure out the response now

josephmancuso:
@girardinsamuel This caused issues with the pip install -e . part of working with Masonite.

Mainly because we cannot installl now any packages that have the masonite.* namespace while working with Masonite in development mode.

This is because installing a package with a masonite.* namespace messes up the import hierarchy.

To get around this, I am moving validation into core and will close down maintaining the validation library as a separate package. I think this solves a few issues as well relating to database rules and maintaining a separate library

src/masonite/dot/dot.py
```diff
@@ -0,0 +1,169 @@
+import pydoc
+import inspect
+from backpack import Collection as collect
+
+
+class Dot:
```
girardinsamuel, line 6: I guess we already have this implementation in masonite.utils.structures 🤔 ?

src/masonite/validation/Validator.py
```diff
@@ -0,0 +1,1329 @@
...
+from masonite.dot import Dot as DictDot
+import inspect
+import re
+import hashlib
+import requests
+import os
+import mimetypes
+
+
+class BaseValidation:
```
girardinsamuel, line 12: So you completely integrated the masonite-validaation package inside Masonite ?
It's totally ok with me 👍, it will be easier to maintain and do more complex stuff

setup.py
```diff
@@ -42,6 +42,7 @@
         "hashids>=1.3,<1.4",
         "cryptography>=3.3.1,<4.0",
         "tldextract>=2.2,<2.3",
+        "masonite-validation>=4.0b1",
```
girardinsamuel, line 45: So is it needed finally ? If all masonite-validation code has been integrated into core ?

josephmancuso: Good point. I'll remove this

src/masonite/dot/dot.py
josephmancuso: I think you're right lol

src/masonite/validation/Validator.py
josephmancuso: yup :)

josephmancuso:
Merging this but the sessions are breaking which we should hopefully fix in the sessions PR

PR #120, opened 2021-06-30
improve sessions

### LINKED ISSUE #119: Session data is not persisted between subsequent requests
I am not talking about flash messages but about session data added with `set()` method.
When doing `request.app.make("session").set("test", "hello")` in a view and getting an other view in the next request
`request.app.make("session").all()` is empty...

It's like every session data is treated as flash messages. I guess the reason is that session is built from cookies and that session is not correctly "hydrated" from the cookies now they are in request and response ??

girardinsamuel:
Try to fix #119 and add flash/reflash/now system as in Laravel

Looks like it's working by playing with it (at least when inspecting the cookie), session helper is offset by one because cookies are deleted at right time but response is already rendered.
I will continue to play with it to fix this

## Fixes
- cookie deletion was not right: I updated delete_cookie() so that cookie are really deleted by expiring them correctly. Also deleted cookies must be sent back to browser to be removed

## Improvments
- add `pull()` method on session
- add `flush()` method on session
- add `push()` method on session

## Changes
Update flash message system based on Laravel ones. Flash messages will be available during the next subsequent request EVEN if they are not used/pulled by `get_flash_messages()` method. It's a better way of doing it IMO. It will also allow methods such as reflash, now, keep to make flash message available during one more request.


src/masonite/cookies/CookieJar.py
```diff
@@ -67,7 +67,7 @@ def load(self, cookie_string):
 
     def render_response(self):
         cookies = []
-        for name, cookie in self.all_added().items():
+        for name, cookie in {**self.deleted_cookies, **self.all_added()}.items():
```
josephmancuso, line 70: Great catch

src/masonite/foundation/response_handler.py
```diff
@@ -35,6 +35,8 @@ def response_handler(environ, start_response):
     to next.
     """
 
+    application.make("session").save()
```
josephmancuso, line 38: Is this the best place for this? Is there any other place we could put this?

src/masonite/providers/HelpersProvider.py
```diff
@@ -17,6 +17,7 @@ def boot(self):
         self.application.make("view").share(
             {
                 "request": lambda: request,
+                "session": lambda: request.app.make("session"),
                 "auth": request.user,
```
josephmancuso, line 21: Just noticed but does this also need to be wrapped in a lambda?

girardinsamuel: I am not sure... maybe to have an up to date session instance ? 
Maybe it's not needed and in a view we could simply use `session.get("key")ˋ

src/masonite/foundation/response_handler.py
girardinsamuel: I agree it might not be the best place. It needs to happen before building response because, at this time we will loop on cookies that should be sent with response and "old flash messages stored in cookies" should have been deleted first. But maybe it can take place in a kind of session handler middleware. 
It's the same session for loading session from request : it could be placed somewhere else...

josephmancuso:
Took a deeper look into this and it looks like a good solution here

josephmancuso:
I took a deeper pen and paper thought on this.

After some doodles and drawing i think the sessions should work like this:

1. What is currently the driver classes should ONLY be used for "starting" and "saving" the session. **This should happen in the before middleware**
2. What is currently the manager class should only be responsible for STORING STATE OF getting, setting, flashing, deleting, etc.
3. When you set a session key and value it should be set on what is currently the session manager class.
4. The session manager class should now act similar to how we built the blueprint Table class (how it maintained a `self.added_constraints`, `self.removed_indexes`, `self.changed_columns`, etc)
5. Then in the **after middleware**, what is currently the session manager class should pass along the state to the cookie driver to "save" it.
6. The cookie driver will then take all the `self.added` and add them to the response (`response.cookie(key, value)`), it should take all the deleted cookies and delete them from the response, (`response.delete_cookie('key')`). It should take the flash data and save them also to the response. The only way we can really identify a flashed value is by the `f_` prefix. `response.key('f_key{}')`.

In summary, we are taking fetching the session and maintaining the state of sessions AWAY from the request and response. and the session will only pull from the request when we start the session and then save to the response when we save the session.

@girardinsamuel what are your thoughts?

josephmancuso:
Above explanation is done and added to this PR

PR #121, opened 2021-07-11
Improve sessions 2

josephmancuso:
Improves #120 

TODO:

- [x] Start
- [x] Save
- [x] Delete
- [x] Pull
- [x] Flush
- [x] Flash
- [x] Has
- [x] Decrement
- [x] Increment


josephmancuso:
@girardinsamuel Here is basically a better version of #120 i think

josephmancuso:
@girardinsamuel I'll keep this open until monday night but the jist is:

What used to be the session manager is now realy the session store that handles the session state throughout the request.

What was the cookie driver is now only responsible for starting and saving the session.

For the cookie driver this means that starting means loading the data from the request into the session class and `saving` means adding or deleting cookies in the response.

For a redis cache this might mean something differently. Either way it should now be really easy to add new session drivers as now they only have to do 2 things: start and save

josephmancuso:
Still need to clean up any now used code

PR #125, opened 2021-07-23
Add with_errors() to response

### LINKED ISSUE #117: Need to make sure several helpers work
There are some quality of live features we need to make sure work:

* Need to make sure you can get old input. This was done in M3 using the `old('input')` helper.
* Need to make sure we can flash data to the session from the request class. This was done doing something like self.with_errors()
* Need to flash the input back to the session. This was used with the old helper and looked something like request.with_input()

@girardinsamuel This might actually need to be done on the response class and now look something like this:

```python
return response.back().with_errors().with_input()
```

TODO:

- [x] old()
- [x] with_input()
- [x] with_errors()

girardinsamuel:
Fixes #117


girardinsamuel:
Not really satisfying but it's working now. The difficulty was that view was returning data and not response instance.
I think it should return response instance and getting data should be done in the response handler at the very end of the cycle.

This way we can always chain back(), redirect() with_input(), with_errors() ensuring we always have a valid response object.

src/masonite/providers/RouteProvider.py
```diff
@@ -39,7 +42,13 @@ def boot(self):
             exception = None
 
             try:
-                response.view(route.get_response(self.application))
+                data = route.get_response(self.application)
```
girardinsamuel, line 45: I am not really satisfied with that. It feels already like a patch...

src/masonite/providers/RouteProvider.py
```diff
@@ -39,7 +42,13 @@ def boot(self):
...
+                if isinstance(data, Response) or (
+                    isclass(data) and issubclass(data, ResponseFacade)
```
josephmancuso, line 47: This looks really messy too .. Maybe we check for a method or attribute on the object?

josephmancuso (changes requested):

PR #126, opened 2021-07-23
Feat/123: add arguments parsing to route middleware

girardinsamuel:


josephmancuso:
Took  a little simpler approach in another PR

Issue #123, opened 2021-07-14
Middleware can no longer accept arguments


josephmancuso:
Somewhere along the way we missed the requirement that middleware should accept values like `auth:parameter`. This is used for route middleware

girardinsamuel:
Could you add an example of usage of a middleware with such an argument ?

josephmancuso:
@girardinsamuel 

```python
ROUTES = [
    Route.get(..).middleware('middleware:argument1,argument2')
]
```

josephmancuso:
V3 had this functionality. Only place we used it internally was to set the guard (`guard:api`) which changes the guard for that route

PR #127, opened 2021-07-24
Add gate and policy feature

girardinsamuel:
This PR is aiming to add an authorization feature through Gate and Policy system (as in Laravel).

### Gates
You needs to define `Gates` in a service provider : 
```python
Gate.define("create-post", lambda user: user.is_admin)
Gate.define("update-post", lambda user, post: post.user_id == user.id)
```

Then anywhere (often in a controller) you can check those gates against the current authenticated user (from request):
```python
if not Gate.allows("create-post"):
    return response.redirect("/")

# ...
post = Post.find(request.input("id"))
if Gate.denies("update-post", post):
    return response.redirect("/")
```
An exception can be raised if needed (and will be rendered as an HTTP response)
```python
Gate.authorize("update-post", post)
# if we reach this part user is authorized
# else an exception has been raised and be rendered as a 403 with a content "Action not authorized".
```

The authorization response can be analyzed :
```python
response = Gate.inspect("update-post", post)
if response.allowed():
     # do something
else:
     # not authorized and we can access message
     Session.flash("errors", response.message())
```

A `Authorizes` model mixin can be added to User model to allow handy permissions checking:
```python
class User(Model, Authenticates, Authorizes):
     # ....

user.can("delete-post", post)
user.cant("access-admin")
user.can_any(["delete-post", "force-delete-post"], post)
```

Before and After hooks are ran during each permission check. A before hook can be added like this : 
```python
# here admin users will always be authorized whatever is happening after
Gate.before(lambda user, permission : user.role == "admin")
```

After hook is working the same excepts that it will receive the authorization result (but if after callback returns something it will take precedance over the permission result check
```python
Gate.after(lambda user, permission, result : user.role == "admin")
```

### Policies

You can create a policy with `python craft policy AccessAdmin` or create a model policy (with a set of predefined gates) with `python craft policy Post --model`.

Then somewhere in a provider (as for defining gates) you should register the policies and bind them with a model :
```python
Gate.register_policies(
    [Post, PostPolicy],
    [User, UserPolicy]
)
```

A policy is looking like
```python
from masonite.authorization import Policy

class PostPolicy(Policy):
    def create(self, user):
        return user.email == "idmann509@gmail.com"

    def view(self, user, instance):
        return True

    def update(self, user, instance):
        return user.id == instance.user_id
```

Then we can check permissions :
```python
post = Post.find(1)

Gate.allows("update", post)
Gate.denies("view", post)
Gate.allows("create", Post) # here we need this to tell which policy is used (it could also be a path to the model => to implement)

user = User.find(2)
Gate.allows("force_delete", user)
Gate.allows("view_any", User)
```

## Improvements (other PR later)
- auto-load policies (low priority)
- define gates with class@method instead of callables  (low priority)
- add an AuthorizeMiddleware `Route.patch("posts/@post/").middleware("can:update,post") or Route.get(...).middleware("can:create,Post")` => would require #123 first.

## Changes in this PR
- add `authorization` feature described above
- add `withExceptionsHandling()` helper to enable exception handling as in real app flow during a unit test
```python
def test_something(self):
     self.withExceptionHandling()
     self.get("/something").assertContains("test") # ==> here if an exception is raised it will be handled by the exception handler
```
- add the ability to render exceptions as HTTP response inside the exception handler. For this, an exception just needs to have a `get_response()` method. If such an exception is raised, the exception handler will try to render it. Here it is used when checking permission with `authorize()` (which raises an `AuthorizationException` if not authorized) : in a controller this will returns a 403 HTTP response with a message.


src/masonite/exceptions/ExceptionHandler.py
```diff
@@ -38,6 +38,9 @@ def handle(self, exception):
                 f"{exception.__class__.__name__}Handler"
             ).handle(exception)
 
+        if getattr(exception, "get_response"):
+            return response.view(exception.get_response())
```
girardinsamuel, line 42: I am not sure how it could be done here, because I would like the status code to be controller by the exception.
If I am doing ```return response.view(exception)```, exception.get_response() will be called but status code will be 200.
Here the `AuthorizationException` is implementing get_response which is returning something like `"Not authorized", 403`.

src/masonite/authorization/AuthorizesRequest.py
```diff
@@ -0,0 +1,5 @@
+class AuthorizesRequest:
+    def authorize(self, permission, *args):
+        from ..facades import Gate
```
josephmancuso, line 3: Does this need to be a lazy import? Can we import this at the top?

josephmancuso (commented):
Looks good so far. Looks very complete as well can't wait to test it out

src/masonite/exceptions/ExceptionHandler.py
josephmancuso: Couldn't you do something like 

```python
response.view(exception.get_response(), exception.get_status())
```

girardinsamuel: I'll do that !

src/masonite/authorization/AuthorizesRequest.py
girardinsamuel: Maybe yes, I will check if there are no issues while doing this

josephmancuso:
Let's get these merge conflicts fixed after i just merged a bunch of PRs

josephmancuso (approved):

PR #128, opened 2021-08-01
Feature/123

### LINKED ISSUE #123: Middleware can no longer accept arguments

josephmancuso:
Closes #123

### DIFF
--- a/src/masonite/middleware/middleware_capsule.py
+++ b/src/masonite/middleware/middleware_capsule.py
@@ -40,14 +40,16 @@ class MiddlewareCapsule:
 
     def run_route_middleware(self, middlewares, request, response, callback="before"):
         for middleware in middlewares:
-            if ':' in middleware:
-                middleware_to_run, arguments = middleware.split(':')
+            if ":" in middleware:
+                middleware_to_run, arguments = middleware.split(":")
                 arguments = (arguments,)
             else:
                 middleware_to_run = middleware
                 arguments = ()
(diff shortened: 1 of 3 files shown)


src/masonite/providers/RouteProvider.py
```diff
@@ -33,32 +33,23 @@ def boot(self):
         )
         if route:
             request.load_params(route.extract_parameters(request.get_path()))
-            Pipeline(request, response).through(
-                self.application.make("middleware").get_route_middleware(
-                    route.list_middleware
-                ),
-                handler="before",
+            self.application.make("middleware").run_route_middleware(
```
girardinsamuel, line 36: I like this !!!

src/masonite/middleware/middleware_capsule.py
```diff
@@ -37,3 +37,19 @@ def get_route_middleware(self, keys=None):
...
+                arguments = (arguments,)
+            else:
+                middleware_to_run = middleware
+                arguments = ()
+            route_middlewares = self.get_route_middleware([middleware_to_run])
+            for route_middleware in route_middlewares:
+                middleware_response = getattr(route_middleware(), callback)(
+                    request, response, *arguments
+                )
+                if middleware_response != request:
```
girardinsamuel, line 54: Route middlewares returns request when everything is okay ? I was thinking they would return `response` ? 🤔

PR #133, opened 2021-10-03
Fix failing tests

### LINKED ISSUE #132: Fix failing tests
Tests are failing right now since `response` refactoring.
In Webguard.py there is a mismatch between setting cookies on request and on response...

attempt, attempt_by_id and logout method set and remove cookies both on request and response. I guess it should be  always on request ?

@josephmancuso what do you think ? I am not sure here !

girardinsamuel:
Fix #132


girardinsamuel:
I am merging this to resolve unit tests currently failing it's a really minor changes. Feel free to review it when you have time.
We can revert this if not okay.

PR #140, opened 2021-10-05
Check and add Python 3.10 support

### LINKED ISSUE #139: Add Python 3.10 support
- [x] Check python 3.10 is okay
- [x] Add python 3.10 in github CI test matrix
- [x] Add python 3.10 support in package metadata

girardinsamuel:
Fixes #139.

It just works ! (based on the unit tests)

There is one deprecation warning but Python 3.12, let us a bit of time...

### DIFF
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


josephmancuso (approved):
Masonite 4 is going to drop support for Python 3.6. I only want to support the last 4 Python releases. Will help keep Masonite moving forward

PR #151, opened 2021-10-08
WIP: Ease package creation

### LINKED ISSUE #79: Add utilities to ease creating packages
I started this for M3 and I would like to migrate it to M4. Two possibilities: either in M4 core or as an external package. For me it makes sense to have this into the core.

Different approachs can be used here to provide utilities to developers wanting to create a package. In this issue I am going down the road of the provider approach. This is a first draft proposal to explain how this could work with M4.
It's inspired from https://github.com/spatie/laravel-package-tools

Of course this might not be the best approach but that's a try 😅 
 
## Specifications

M4 could provide a `PackageProvider` that you can use in your packages, to easily:
- register package configuration (and publish it if you want to override it)
- register routes (and publish them if you want to override them)
- register views (and publish them if you want to override them)
- register migrations (and publish them if you want to override them)
- register tests assertions
- register assets (that the local project can find and serve => register a location.) (and publish them if you want to override them) 
- register commands

All files could be published by default under a configurable "vendor/package-name" namespace in locations defined in Kernel.py of project. So we would use what is defined in the container to build correct locations.

This is how it could look like:

```python
from masonite.providers import PackageProvider

package_root_path = "path/to/package/root"

class MyPackageProvider(PackageProvider):

    def configure(self):
        self.name("my-package")
        self.add_config("test")
        self.add_command(MyCommand())
        self.add_routes("admin", "public", tag="my-routes")
        self.add_migrations("create_my_table", "create_my_other_package_table", tag="db")
        self.add_assets(
                {
                    "package.js": "app.js",
                    "package.css": None,
                    "sub/admin.js": "admin.js",
                }
       )
       seld.add_tests_assertions(CustomTestResponse)
       self.add_command(OtherCommand(), "DemoCommand")
```

M3 `PackageProvider` code for reference https://github.com/MasoniteFramework/masonite/pull/411/files#diff-26aa7ccb08af6962c9fcb17624445040dc5fefb313ff32a4ab939e8e2c7c0ed9 that could be used to rewrite that from scratch for M4. Here the PackageProvider, call configure method behind the scenes in the register() method. And add stuff in the boot method also.

## Small details in API
- Naming of methods could also be: self.routes(), self.register_routes(),
- Each time we could add one element or several : what should be preferred ? add_commands(Command1, Command2), add_command(Command1)
- For each kind of element we want to register it's important IMO to distinguish that the element is registered and available in the project but not copied over. If we wanted to override or extend the element then we can publish it to copy it into our projects. Example: I registered some routes from an external package. Those don't appear in my route files and it's ok. I want to customize route names or more, I will publish those routes into my project. This PackageProvider should handle both: register stuff in register() and in boot() make stuff available to be published.

girardinsamuel:
Try some ideas to fix #79.

API in progress and waiting for PR #155, #154, #152.

If package layout is:
```
src
   masonite
       test_package
          config
             test.py
          templates
             my_view.html
          assets
             test.js
          migrations
            create_some_table.py
          routes
            api.py
            web.py
```

```python
class MyTestPackageProvider(PackageProvider):
    def configure(self):
        (
            self.root("src/masonite/test_package")
            .name("test_package")
            .add_config("config/test.py", publish=True)
            .add_views("templates", publish=True)
            .add_commands(Command1(), Command2())
            .add_migrations("migrations/create_some_table.py")
            .add_assets("assets")
            .add_routes("routes/api.py", "routes/web.py")
        )
```
With this the commands are added, the configuration is loaded the views can be used with namespace `view("test_package:my_view")`, migrations will be published, assets will be published and routes will be added to in router.

To publish all files to project:
```
$ python craft package:publish test_package
````

To publish only config
```
$ python craft package:publish test_package --resources config
```

To publish only config and views...
```
$ python craft package:publish test_package --resources config,views
```

Want to see what is going to be published ? Use `--dry` option:
image

Here we have long paths `tests/integrations/..` because of the test project inside M4 repo but it will be more clear and easier with a normal project.

### CODE REVIEW

**Comment by girardinsamuel** [comment]
> Hi @girardinsamuel As suggestion for the api, is possible using fluent api during the "add" methods of each part of new package? Something like:
> 
> ```
> self.root(...)
> .name(...)
> .config(...)
> .commands(..)
> .migrations(...)
> ...
> ```
> 
> Just a simple api to define each part. No more important to core funtionality.

You mean removing the `add_` prefix for each method ? Yes why not 😉 !

**Comment by girardinsamuel** [comment]
there is one test failing. at some point in the tests suite routes are overriden and the initial routes shared by the package provider are overriden making the test fail because the router don't find the routes. When the test fail the router is containing ~229 routes with the same url / and without name....
Running this test independently works

**Comment by josephmancuso** [comment]
I tried taking a look at this and couldn't figure out the issue ... There is another issue i found though which is that in the testCase class we are adding routes everytime the class is setup:

```python
class TestCase(TestCase):
    def setUp(self):
        super().setUp()
        self.addRoutes(
            Route.get("/", "WelcomeController@show"),
            Route.post("/", "WelcomeController@show"),
        )
```

for some reason this is causing those 2 routes to get added on every single test. So we end up with like 800 routes with the `/` url

josephmancuso:
Just removed the route stuff for now. We can revisit it in the future

PR #158, opened 2021-10-13
Boot providers during tests

### LINKED ISSUE #153: code in providers boot() not called during the tests
That's not a big deal as more often it is used as part of the lifecycle request but I can see some use cases:
- during unit tests
- in somme console commands

For example, I was trying to test View facade to render a view. We might have to do this in a console command or to assess behaviour in a unit test, and here as boot() methods of providers are not called, some stuff are not shared in the view that are usually shared (session, request,...) and that can lead to errors.

Should be boot providers during tests if we are not making a request ?
What's your opinon on that ?

girardinsamuel:
Fixes #153.

Here I don't load the RouteProvider during tests as we don't have a CSRF token yet. Of course it will be loaded when doing a request such as `self.get()` or an other method.

I added a test on view which is testing that helpers are correctly shared in views. They are shared during boot of ViewProvider, so this a proof that it is working 😉

### DIFF
--- a/src/masonite/providers/SessionProvider.py
+++ b/src/masonite/providers/SessionProvider.py
@@ -13,10 +13,9 @@ class SessionProvider(Provider):
         session = Session(self.application).set_configuration(config("session.drivers"))
         session.add_driver("cookie", CookieDriver(self.application))
         self.application.bind("session", session)
-        self.application.make("view").share({"old": self.old})
 
     def boot(self):
-        pass
+        self.application.make("view").share({"old": self.old})
 
     def old(self, key):
         return self.application.make("session").get(key) or ""
--- a/src/masonite/providers/SessionProvider.py
+++ b/src/masonite/providers/SessionProvider.py
@@ -13,9 +13,10 @@ class SessionProvider(Provider):
         session = Session(self.application).set_configuration(config("session.drivers"))
         session.add_driver("cookie", CookieDriver(self.application))
         self.application.bind("session", session)
+        self.application.make("view").share({"old": self.old})
 
     def boot(self):
-        self.application.make("view").share({"old": self.old})
+        pass
 
     def old(self, key):
         return self.application.make("session").get(key) or ""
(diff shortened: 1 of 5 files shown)


src/masonite/providers/SessionProvider.py
```diff
@@ -13,10 +13,9 @@ def register(self):
         session = Session(self.application).set_configuration(config("session.drivers"))
         session.add_driver("cookie", CookieDriver(self.application))
         self.application.bind("session", session)
-        self.application.make("view").share({"old": self.old})
 
     def boot(self):
-        pass
+        self.application.make("view").share({"old": self.old})
```
josephmancuso, line 18: why is this moved into the boot?

girardinsamuel: I don't know lol I was thinking that it was the correct place to define view sharing...

girardinsamuel: do you want to revert this ?

josephmancuso: I don't think it is right? Well it depends. In this case the old method is not called until you call it in your templates. 

In this case it will make each request slower because it has to register this old method every request. 

If we keep it in the register method it only gets registered once and it should still work

src/masonite/tests/TestCase.py
```diff
@@ -25,6 +27,18 @@ def setUp(self):
             self.startTestRun()
         self.withoutCsrf()
         self._exception_handling = False
+        # boot providers as they won't not be loaded if the test is not doing a request
+        self.application.bind("environ", generate_wsgi())
+        try:
+            for provider in self.application.get_providers():
+                # if no request is made we don't need RouteProvider, and we can't load it anyway
+                # because we don't have created a CSRF token yet
+                if not isinstance(provider, RouteProvider):
```
josephmancuso, line 36: Also i thought we get around the csrf checks by setting a cookie and the csrf input?

josephmancuso: hmm not sure what the best approach is but if this works i'll merge it in and we can revisit it later

josephmancuso (approved):

PR #160, opened 2021-10-16
Added Notification Facade

alitokmakci:
Docs says Notification Facade is available but there is no Facade called Notification.


girardinsamuel (changes requested):
@alitokmakci thanks for this fix ! You just need to also add the import of this facade here
https://github.com/MasoniteFramework/masonite4/blob/master/src/masonite/facades/__init__.py

Thanks !

PR #193, opened 2021-11-06
Add back test case assertions and craft test command

### LINKED ISSUE #192: Add command to create a new test file

girardinsamuel:
Fixes #192
Fixes https://github.com/MasoniteFramework/masonite/issues/456
This PR adds back:
- craft test command
- `captureOutput()` helper in test
- uncomment some commented assertions tests which are working now
- add ability for assertRouteParameterHas() to check param value

`python craft test MyAppTest` will create:

```python


class MyAppTest(TestCase):

    def setUp(self):
        super().setUp()

    def test_something(self):
        self.assertTrue(True)

```

in `tests/unit/test_my_app.py`

Directory can be changed with `python craft test MyFeatureTest --directory tests/features`


src/masonite/stubs/tests/TestCase.py
```diff
@@ -0,0 +1,9 @@
+from masonite.tests import TestCase
```
josephmancuso, line 1: new Masonite projects actually have a base test case inside `tests/TestCase`. this import should be `from tests.TestCase import TestCase`

src/masonite/tests/TestCase.py
```diff
@@ -47,10 +47,14 @@ def tearDown(self):
...
 
     def withExceptionsHandling(self):
-        """Enable for the duration of a test the handling of exception through the exception
+        """Enable for the duration of a test the handling of exceptions through the exception
         handler."""
         self._exception_handling = True
 
+    def withoutExceptionsHandling(self):
+        """Disable handling of exceptions."""
+        self._exception_handling = True
```
josephmancuso, line 56: if this is **without**exception handling should this be setting the attribute to `True`? it makes more sense if this is `False` right?

josephmancuso (changes requested):

src/masonite/tests/TestCase.py
girardinsamuel: Wooops !

src/masonite/stubs/tests/TestCase.py
girardinsamuel: You're right !

PR #216, opened 2021-12-17
Fix back() helper

### LINKED ISSUE #215: Check/Fix back view helper
I recreate the issue in M4 to be sure we address it in M4 : https://github.com/MasoniteFramework/masonite/issues/382

# Description of the issue (correct me if wrong)
When I am calling the view including `{{ back() }}` with some query parameters, those are not included inside the "back generated input".

# Elements to solve it

this helper use `request.get_path()` which do not contain query params.

We could add to request class: 
```python
# existing
def get_path(self):
    return self.environ.get("PATH_INFO")

# new
def get_path_with_query(self):
    return self.environ.get("PATH_INFO") + "?" + self.environ.get("QUERY_STRING")
```

And then in the helper:
```python
# HelpersProvider.py

"back": lambda url=request.get_path_with_query(): (
    Markup(f"<input type='hidden' name='__back' value='{url}' />")
),
```

With this when visiting `http://localhost:8000/test?key=val` `{{ back() }}` will generate: 
```html
<input type="hidden" name="__back" value="/test?key=val">
```

girardinsamuel:
Fixes #215.

To review carefully as I am not sure of the implications. all tests are passing though 😅

### DIFF
--- a/src/masonite/providers/HelpersProvider.py
+++ b/src/masonite/providers/HelpersProvider.py
@@ -25,7 +25,7 @@ class HelpersProvider(Provider):
                 "session": lambda: request.app.make("session"),
                 "auth": request.user,
                 "cookie": request.cookie,
-                "back": lambda url=request.get_path(): (
+                "back": lambda url=request.get_path_with_query(): (
                     Markup(f"<input type='hidden' name='__back' value='{url}' />")
                 ),
                 "asset": urls_helper.asset,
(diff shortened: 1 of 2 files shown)


josephmancuso (approved):
Looks right 🤷 I can test it more tonight

archibien:
Well let's merge it in then !

girardinsamuel:
@josephmancuso did you find time to test it more ? Maybe we can just merge it in !

josephmancuso (approved):

josephmancuso:
I trust you :)

PR #218, opened 2021-12-21
Add API features

josephmancuso:
The first iteration of API improvements in M4 will be pretty small. Just the requirements.

So to build an API using these features is something like this:

## Step 1

You can create a normal controller or create a resource controller:

```python
python craft controller UserController -r
```

## Step 2

Create your route, could be normal routes or you can use the new `resource` method which maps several routes to the matching controller method:

```python
ROUTES = [
    Route.resource('users', 'UserController')
]
```

## Step 3

Add the authentication controller:

```python
python craft api:install
```

This will scaffold out the authentication controller so you can modify it

## Step 4

Call `/api/auth` with your username and password and get back a token which is saved to the users table. You can use this token to make requests.

The JWTAuthenticationMiddleware will verify this token is passed and that it is valid


src/masonite/api/authentication/AuthenticatesTokens.py
```diff
@@ -0,0 +1,12 @@
+import jwt
+
+
+class AuthenticatesTokens:
+
+    __TOKEN_COLUMN__ = "api_token"
```
girardinsamuel, line 6: So we should add this column in the default User model ? 
What would be the type/length of the field ?

src/masonite/api/middleware/JWTAuthenticationMiddleware.py
```diff
@@ -0,0 +1,19 @@
+from masonite.middleware import Middleware
+from masonite.facades import Auth
+
+
+class JWTAuthenticationMiddleware(Middleware):
+    def before(self, request, response):
+        token = request.input("token")
+        if not token:
+            return response.view(
```
girardinsamuel, line 9: Those views will be converted to json if json is requested I presume ? (I guess you fix that few days ago)

src/masonite/stubs/api/AuthenticationController.py
```diff
@@ -0,0 +1,16 @@
...
+from masonite.response import Response
+from masonite.authentication import Auth
+
+
+class AuthenticationController(Controller):
+    def auth(self, auth: Auth, request: Request, response: Response):
+        user = auth.attempt(request.input("username"), request.input("password"))
+
+        if user:
+            return {"data": user.generate_jwt()}
```
girardinsamuel, line 12: here it means that a new JWT token is generated each time we make a request to this controller, replacing the token in user table with the newly generated ?

And what's the lifetime of this token ?

Also should we rename `data` to `token` ?

src/masonite/stubs/api/AuthenticationController.py
```diff
@@ -0,0 +1,16 @@
...
+
+        return response.json(
```
girardinsamuel, line 14: here you returned JSON and in the middleware we return a view. Couldn't we return json directly or am I missing something ?

girardinsamuel (commented):
Made a quick review !

src/masonite/api/authentication/AuthenticatesTokens.py
josephmancuso: @girardinsamuel good question. Would probably just be a varchar 255

src/masonite/api/middleware/JWTAuthenticationMiddleware.py
josephmancuso: yeah the view method does the response conversion

src/masonite/stubs/api/AuthenticationController.py
josephmancuso: Yes. Once you have a token you would not need to hit this endpoint anymore. 

The lifetime of the token will be inside the JWT token which I have to add inside the generate_gwt() method. Somehow I still need to hook this all up to the normal feature convention we have so I have access to the configuration settings in the same way we do with other features

src/masonite/stubs/api/AuthenticationController.py
josephmancuso: we could yeah. The view method detects the input type so if you did `.view({dict})` it will just call `.json`. `.view()` actually just wraps the controller method so that how we handle all the different response types you can return

but I agree it should be consistent here

src/masonite/api/authentication/AuthenticatesTokens.py
girardinsamuel: fine to me !

src/masonite/stubs/api/AuthenticationController.py
girardinsamuel: Yep it would be great. because I feel we would like to define some settings to generate JWT.

src/masonite/stubs/api/AuthenticationController.py
girardinsamuel: yep that's what I was pointing out

PR #219, opened 2021-12-21
Add ORM commands

### LINKED ISSUE #214: craft model should follow models location configuration
`python craft model Post` should create the file at location defined by `models.location` binding.

# Problem
The issue here is that this config is defined in Masonite and the command is coming from masonite-orm.

Here are some solutions to address this issue :

## Approach 1
Should we add an argument in ORMProvider when initializing MakeModelCommand ?

```python
class ORMProvider(Provider):
    """Masonite ORM database provider to be used inside
    Masonite based projects."""

    def __init__(self, application):
        self.application = application

    def register(self):
        self.application.make("commands").add(
            MakeMigrationCommand(),
            MakeSeedCommand(),
            MakeObserverCommand(),
            MigrateCommand(),
            MigrateResetCommand(),
            MakeModelCommand(application.make("models.location")),
            MigrateStatusCommand(),
            MigrateRefreshCommand(),
            MigrateRollbackCommand(),
            SeedRunCommand(),
        ),

    def boot(self):
        pass
```

This would require to edit the MakeModelCommand __init__ method to add this argument and use it when defined and when not defined use the directory option...

## Approach 2
Duplicate this command inside Masonite commands and don't register the command in ORMProvider => then we can do what we want.

## Approach 3
Duplicate this command inside Masonite commands and don't register the command in ORMProvider. But inherit the MakeModelCommand from masoniteorm.

girardinsamuel:
Fixes #214 
For now I added all commands available inside ORMProvider from orm package and then I swapped the `MakeModelCommand` with an internal one. I guess we should do it for some other commands so that models location and migrations location are correctly configured without to pass --models or --directory option each time.

To achieve this I added a way to swap commands in the command capsule by another one at the same key, and then I registered an "Internal ORM Provider" to make the swapping.

Finally end users just see the `ORMProvider` coming from the `masonite` package.

What's left to do :  duplicate other commands that require some location already defined in Masonite -> actually all commands need this.


src/masonite/commands/MakeModelCommand.py
```diff
@@ -0,0 +1,54 @@
+"""New Model Command."""
```
josephmancuso, line 1: I though we would just make a new parameter inside the __init__ and then in Masonite we pass in the default model directory?

src/masonite/orm/providers/InternalORMProvider.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class InternalORMProvider(Provider):
+    """Allow to tweak official ORMPRovider when used inside Masonite."""
+
+    def __init__(self, application):
+        self.application = application
+
+    def register(self):
+        self.application.make("commands").swap(MakeModelCommand(self.application))
```
josephmancuso, line 15: Then in here we do 

```python
MakeModelCommand(self.application, default_model_path=model_path())
```

tests/integrations/config/providers.py
```diff
@@ -47,4 +49,6 @@
     ValidationProvider,
     MyTestPackageProvider,
     AppProvider,
+    ORMProvider,
+    InternalORMProvider,
```
josephmancuso: I don't like 2 different ORM Providers. we should put all of Masonite ORM's commands inside 1 Masonite provider inside Masonite core and modify the commands to accept inputs for anything diff

josephmancuso (changes requested):
I think we can modify this a bit to make it cleaner

src/masonite/commands/CommandCapsule.py
```diff
@@ -8,5 +8,13 @@ def add(self, *commands):
         self.command_application.add_commands(*commands)
         return self
 
+    def swap(self, command):
+        command_name = command.config.name
+        # if command with same name has been registered remove it
+        if self.command_application.find(command_name):
+            # no public API to do this yet
+            del self.command_application.commands._commands[command_name]
+        self.add(command)
```
josephmancuso, line 17: This is a good method we should keep it

tests/integrations/config/providers.py
girardinsamuel: I edited the PR at the same time there are no different providers anymore.

src/masonite/orm/providers/InternalORMProvider.py
girardinsamuel: Yep this would avoid duplication, I will try

girardinsamuel:
It looks really clean to me now ! @josephmancuso Ready for review !
And the cookie-cutter is updated here : https://github.com/MasoniteFramework/cookie-cutter/pull/20

josephmancuso (approved):
Looks good!

PR #220, opened 2021-12-21
Fix using `setRoutes()` or `addRoutes()`

### LINKED ISSUE #190: Add back in the routes feature for the "ease of use" package feature

girardinsamuel:
When `setRoutes` or `addRoutes` method are called during tests, routes will be registered only during the lifetime of the test. This will ease having idempotent and isolated tests.

- Routes registering during tests has been cleaned. We won't end up with 300 routes at the end of the tests anymore
- It fixes #190

### DIFF
--- a/src/masonite/routes/HTTPRoute.py
+++ b/src/masonite/routes/HTTPRoute.py
@@ -38,6 +38,9 @@ class HTTPRoute:
         self.controller_bindings = controller_bindings
         self.compile_route_to_regex()
 
+    def __str__(self):
+        return f"<HttpRoute [{self._name}]: {self.url}>"
+
     def match(self, path, request_method, subdomain=None):
 
         route_math = (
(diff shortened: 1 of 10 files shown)


src/masonite/tests/TestCase.py
```diff
@@ -138,12 +147,12 @@ def fetch(self, route, data=None, method=None):
...
             request.header(name, value)
 
-        route = self.application.make("router").find(route, method)
+        route = self.application.make("router").find(path, method)
         if route:
             return self.application.make("tests.response").build(
                 self.application, request, response, route
             )
-        raise Exception(f"No route found for {route}")
+        raise Exception(f"No route found for url {path}")
```
girardinsamuel, line 155: Instead of printing `No route found for None` we get `No route found for /not-existing`.  `route` which was None in L150 was shadowing `route` arg provided to the function

src/masonite/routes/HTTPRoute.py
```diff
@@ -38,6 +38,9 @@ def __init__(
         self.controller_bindings = controller_bindings
         self.compile_route_to_regex()
 
+    def __str__(self):
```
girardinsamuel, line 41: Really easier to debug to have a clear and unique reprensation of the object in console.

josephmancuso (approved):

PR #221, opened 2021-12-22
Don't override files by default when using "make" commands

girardinsamuel:
All "make" commands now have a force option to avoid overriding the file. A warning is displayed and the command stops.
The user is invited to use `--force` flag to override the file.

Also line_error() was used and sometimes without style="error" so to be more consistent, now all commands use and can access:
```python
self.info()
self.comment()
self.warning()
self.error()
```

### DIFF
--- a/src/masonite/commands/AuthCommand.py
+++ b/src/masonite/commands/AuthCommand.py
@@ -1,10 +1,10 @@
 """Scaffold Auth Command."""
-from cleo import Command
 from distutils.dir_util import copy_tree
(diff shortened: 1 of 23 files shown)


girardinsamuel:
@josephmancuso are you okay with this ? It's just some security. I don't really like when override is not mentioned...can surprise some users !

josephmancuso (approved):

josephmancuso:
Great idea!

